PR #10, opened 2015-09-06
More tests

kain88-de:
This removes some old cruft and adds new tests. The division test is still failing under python3 I'll check tomorrow what the reason might be.

Also I saw that you have several additions defined. A normal one and a special per grid point one. But since we use numpy they are the same in the current implementation. Care if I remove them?

btw this brings coverage to 50%

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: python
+python:
+  - "2.7"
+  - "3.4"
+addons:
+  apt:
+    packages:
+      - gfortran
+      - liblapack-dev
+      - libatlas-dev
+before_install:
+  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
+  - chmod +x miniconda.sh
+  - ./miniconda.sh -b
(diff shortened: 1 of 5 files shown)


kain88-de:
It was late yesterday. I would remove the `__r<op>__` functions. I'm also working a little bit to get the coverage score of at least `core.py` a little bit higher

kain88-de:
> It was late yesterday. I would remove the __r<op>__ functions.

nevermind. a quick google search explained what these operator are for. I'm testing them as well now.

richardjgowers:
Cool, if you've added travis in you should also put up the badges for this [like here](https://raw.githubusercontent.com/MDAnalysis/mdanalysis/develop/README.rst)

kain88-de:
OK this is now testing a lot of the grid object. I haven't added test's for the file formats a lot. This is because I think all of them should be refactored to be simpler and best follow a simple API, especially the openDX file. Expect for the interpolation algorithm all the math stuff in the Grid object should be tested now.

I also refactored the code a little bit to make testing easier and be sure to get the math right.
Could have been I fixed a bug or two with that.

PR #14, opened 2015-09-13
This should container builds in travis-CI

kain88-de:
lets see if this works

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - "2.7"
   - "3.4"
+sudo: false
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


orbeckst:
I thought container builds are the default nowadays, and you'd only get something different with `sudo: True`.

kain88-de:
I got a message about using the old infrastructure on the travis build. The docs say that they are only explicitly activated with the sudo flag

http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

orbeckst:
Hi Max,

> I got a message about using the old infrastructure on the travis build. The docs say that they are only explicitly activated with the sudo flag

Then I obviously mis-remembered. I thought that MDAnalysis runs on the new containers even though the sudo=false is flag is not set. At least that's what I remember from conversations with @richardjgowers and checking the log such as https://travis-ci.org/MDAnalysis/mdanalysis/builds/80046564 says that it is using docker and containers:

 Using worker: worker-linux-docker-bf8ff456.prod.travis-ci.org:travis-linux-3

 This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.
 If you require sudo, add 'sudo: required' to your .travis.yml 
 See http://docs.travis-ci.com/user/workers/container-based-infrastructure/ for details.

Cheers,
Oliver

richardjgowers:
@orbeckst you're not wrong, MDAnalysis defaulted to container based once we got rid of sudo.  I'm not sure why this didn't too, but it's fixed now

PR #31, opened 2016-04-16
Add tempdir module and update test

kain88-de:
This way we remove a dependency and it should help with conda builds.


jbarnoud:
If you get to include tempdir in both MDAnalyssis and GridDataFormats, it may not be the best solution.

orbeckst:
I am going to merge this for now. If we figure out a better solution we can get rid of it again.

Issue #35, opened 2017-07-27
Pymol can't read OpenDX file produced by GridDataFormats


rxu2:
I made a DX file with the following python program:
```
import numpy as np
import gridData

nx= ny= nz = 5
density = np.empty((nx,ny,nz), dtype=float)
for i in range(nx):
    for j in range(ny):
        for k in range(nz):
            density[i,j,k] = i**2 + j**2 + k**2
density /= density.max()

g =gridData.Grid(density, origin=(0, 0, 0), delta=(1,1,1))
g.export('test.dx', 'dx')
```
When I load `test.dx` into pymol (`PyMOL 1.8.4.0 Open-Source` on Linux) with the bash command (`$ pymol test.dx`), I get the following error messages from pymol:

```
 ObjectMapLoadDXFile: Loading from 'test.dx'.
 DXStrToMap: Dimensions: 30 30 30
 DXStrToMap: Origin    0.000    0.000    0.000
 DXStrToMap: Grid    1.000    1.000    1.000
 DXStrToMap: 27000 data points.
ObjectMap-Error: Error reading map
 CmdLoad: "test.dx" loaded as "test".
```

Here is the test.dx file:

```
# OpenDX density file written by gridDataFormats.Grid.export()
# File format: http://opendx.sdsc.edu/docs/html/pages/usrgu068.htm#HDREDF
# Data are embedded in the header and tied to the grid positions.
# Data is written in C array order: In grid[x,y,z] the axis z is fastest
# varying, then y, then finally x, i.e. z is the innermost loop.
# (Note: the VMD dx-reader chokes on comments below this line)
object 1 class gridpositions counts  5 5 5
origin 0.000000 0.000000 0.000000
delta  1.000000 0.000000 0.000000
delta  0.000000 1.000000 0.000000
delta  0.000000 0.000000 1.000000
object 2 class gridconnections counts  5 5 5
object 3 class array type float rank 0 items 125 data follows
0.0	0.0208333333333	0.0833333333333	
0.1875	0.333333333333	0.0208333333333	
0.0416666666667	0.104166666667	0.208333333333	
0.354166666667	0.0833333333333	0.104166666667	
0.166666666667	0.270833333333	0.416666666667	
0.1875	0.208333333333	0.270833333333	
0.375	0.520833333333	0.333333333333	
0.354166666667	0.416666666667	0.520833333333	
0.666666666667	0.0208333333333	0.0416666666667	
[...]

PR #36, opened 2017-12-19
set DX array type

orbeckst:
Issue #35 (and MDAnalysis/mdanalysis#1725) report that DX files written by gridDataFormats cannot be read by PyMOL because PyMOL (wrongly) *hard codes* the expected array type as "double". 

This PR adds a `type` kwarg to `Grid.export()` so that users now have the option to set the output data type to "double" if they so choose; the data are cast to the closest numpy.dtype.

By default, the numpy.dtype of the underlying array is used to set the closest matching DX data type.

### DIFF
--- a/gridData/OpenDX.py
+++ b/gridData/OpenDX.py
@@ -377,7 +377,7 @@ class field(DXclass):
             components = dict(positions=None,connections=None,data=None)
         if comments is None:
             comments = ['OpenDX written by gridData.OpenDX',
-                        'from http://github.com/orbeckst/GridDataFormats']
+                        'from https://github.com/MDAnalysis/GridDataFormats']
         elif type(comments) is not list:
             comments = [str(comments)]
         self.id = classid       # can be an arbitrary string
--- a/gridData/OpenDX.py
+++ b/gridData/OpenDX.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2009-2014 Oliver Beckstein <orbeckst@gmail.com>
 # Released under the GNU Lesser General Public License, version 3 or later.
 
-r""":mod:`OpenDX` --- routines to read and write simple OpenDX files
+r""":mod:`~gridData.OpenDX` --- routines to read and write simple OpenDX files
 ================================================================
 
 The OpenDX format for multi-dimensional grid data. OpenDX is a free
@@ -18,6 +18,66 @@ If you want to build a dx object from your data you can either use the
 convenient :class:`~gridData.core.Grid` class from the top level
 module (:class:`gridData.Grid`) or see the lower-level methods
 described below.
+
+
+Reading and writing OpenDX files
+--------------------------------
+
+If you have OpenDX files from other software and you just want to
+**read** it into a Python array then you do not really need to use the
+interface in :mod:`gridData.OpenDX`: just use
+:class:`~gridData.core.Grid` and load the file::
+
+  from gridData import Grid
+  g = Grid("data.dx")
+
+This should work for files produced by common visualization programs
+(VMD_, PyMOL_, Chimera_). The documentation for :mod:`gridData` tells
+you more about what to do with the :class:`~gridData.core.Grid`
+object.
+
+If you want to **write** an OpenDX file then you just use the
+:meth:`gridData.core.Grid.export` method with `file_format="dx"` (or
+just use a filename with extension ".dx")::
+
+  g.export("data.dx")
+
+However, some visualization programs do not implement full OpenDX
+specifications and only read very specific, "OpenDX-like"
+files. :mod:`gridData.OpenDX` tries to be compatible with these
+formats. However, sometimes additional help is needed to write an
+OpenDX file that can be read by a specific software, as described
+below:
+
+Known issues for writing OpenDX files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* PyMOL_ requires OpenDX files with the type specification "double" in
+  the `class array` section (see issue `#35`_). By default (since
+  release 0.4.0), the type is set to the one that most closely
+  approximates the dtype of the numpy array :attr:`Grid.grid`, which
+  holds all data. This is often :class:`numpy.float64`, which will
+  create an OpenDX type "double", which PyMOL will read.
+
+  However, if you want to *force* a specific OpenDX type (such as
+  "float" or "double", see :attr:`gridData.OpenDX.array.dx_types` for
+  available values) then you can use the ``type`` keyword argument::
+
+    g.export("for_pymol.dx", type="double")
+
+  If you always want to be able to read OpenDX files with PyMOL, it is
+  suggested to always export with ``type="double"``.
+
+  .. versionadded:: 0.4.0
+
+
+
+.. _VMD: http://www.ks.uiuc.edu/Research/vmd/
+.. _PyMOL: http://www.pymol.org/
+.. _Chimera: https://www.cgl.ucsf.edu/chimera/
+.. _`#35`: https://github.com/MDAnalysis/GridDataFormats/issues/35
+
+
 
 
 Building a dx object from a numpy array ``A``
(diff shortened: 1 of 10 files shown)


gridData/OpenDX.py
```diff
@@ -182,18 +192,118 @@ def write(self,file):
...
+        "unsigned int": "uint32",
+        "int": "int32",
+        "signed int": "int32",
+        "unsigned int": "uint64", # not explicit in DX, for compatibility
+        "int": "int64",           # not explicit in DX, for compatibility
+        # "hyper",                # ?
+        "float": "float32",       # default
+        "double": "float64",
+        # "string" not automatically supported
+    }
```
kain88-de: are they exact copies of each other only reverted? Wouldn't it be better to use `dx_types = {v: k for v,k in six.iteritems(np_types)}` so that we only have one list to take care of?

gridData/core.py
```diff
@@ -363,10 +381,11 @@ def _export_python(self, filename):
         with open(filename, 'wb') as f:
             cPickle.dump(data, f, cPickle.HIGHEST_PROTOCOL)
 
-    def _export_dx(self, filename):
+    def _export_dx(self, filename, **kwargs):
```
kain88-de: why not use `type=none` instead of undocumented `**kwargs`?

gridData/OpenDX.py
```diff
@@ -182,18 +192,118 @@ def write(self,file):
...
+
+    def __init__(self, classid, array=None, type=None, **kwargs):
+        """
+        Parameters
+        ----------
+        classid : int
+        array : array_like
+        type : str (optional)
```
kain88-de: I would prefer to name this `dtype` for **D**ata**Type** following the numpy naming. That way we also avoid clashing with the buildin `type`.

gridData/OpenDX.py
orbeckst: No, they are not inverted, there's not a 1-to-1 correspondence.

gridData/OpenDX.py
orbeckst: dtype is misleading because they are not numpy.dtypes. They are the specific labels in the DX file format specification. But "dxtype" seemed clunky and I didn't see a problem in shadowing the builtin type if it was clearer to the user. I also did not want to use dxtype as argument for export because if we ever want to use the data type for other formats then we can parse type appropriately.

gridData/core.py
orbeckst: changed

gridData/OpenDX.py
```diff
@@ -182,18 +192,118 @@ def write(self,file):
...
+        "int64": "int",           # not explicit in DX, for compatibility
+        # "hyper",                # ?
+        "float32": "float",       # default
+        "float64": "double",
+        "float16": "float",       # float16 not available in DX, use float
+        # numpy "float128 not available, raise error
+        # "string" not automatically supported
+    }
+    dx_types = {
+        "byte": "uint8",
```
kain88-de: is unsigned the default in DX files?

orbeckst: no, float

gridData/core.py
```diff
@@ -363,10 +381,11 @@ def _export_python(self, filename):
         with open(filename, 'wb') as f:
             cPickle.dump(data, f, cPickle.HIGHEST_PROTOCOL)
 
-    def _export_dx(self, filename):
+    def _export_dx(self, filename, type=None, **kwargs):
```
kain88-de: kwargs can be removed now.

gridData/core.py
```diff
@@ -346,11 +346,29 @@ def export(self, filename, file_format=None):
...
+
+        .. _`#35`: https://github.com/MDAnalysis/GridDataFormats/issues/35
+
         """
         exporter = self._get_exporter(filename, file_format=file_format)
-        exporter(filename)
+        exporter(filename, type=type)
 
-    def _export_python(self, filename):
+    def _export_python(self, filename, **kwargs):
```
kain88-de: where is `kwargs` used in the function? Wouldn't explicit named args be better?

gridData/core.py
orbeckst: Technically yes; practically it says that all `_export_*()` methods need to be able to take all kwargs and only process the ones that are needed. 

That's also the reason for `_export_python()` just using `**kwargs` and not doing anything with it.

gridData/core.py
orbeckst: See https://github.com/MDAnalysis/GridDataFormats/pull/36#discussion_r157713756

I could add `type=None` and then ignore it but it seems clearer (to me, at least...) that `**kwargs` means that the method does not know what it will get and it will just take what it needs.

orbeckst: I added an explanatory comment in cd3fdb051b86d3a2559461411cbcec60f98102d7

orbeckst:
I have to get some sleep... if you want to merge then just squash everything. Thanks for reviewing.

gridData/OpenDX.py
kain88-de: Sorry I meant is a byte really automatically unsigned?

orbeckst:
Yes, that's my interpretation of the description of the allowed values on the page for the DX format ---see the link somewhere in the PR (sorry, can't get at it easily from mobile). 

> Am Dec 19, 2017 um 07:15 schrieb Max Linke <notifications@github.com>:
> 
> @kain88-de commented on this pull request.
> 
> Sorry I meant is a byte really automatically unsigned?
> 
> 
>

gridData/OpenDX.py
```diff
@@ -182,18 +192,118 @@ def write(self,file):
...
+        "unsigned byte": "uint8",
+        "signed byte": "int8",
+        "unsigned short": "uint16",
+        "short": "int16",
+        "signed short": "int16",
+        "unsigned int": "uint32",
+        "int": "int32",
+        "signed int": "int32",
+        "unsigned int": "uint64", # not explicit in DX, for compatibility
+        "int": "int64",           # not explicit in DX, for compatibility
```
kain88-de: int appears twice in this list. We can only map to a 32 or 64 bit one. Given the age of the format my guess is to use 32bit.

gridData/OpenDX.py
kain88-de: for compatibility we should here map `char` to an `uint8`. 

https://web.archive.org/web/20080808140524/http://opendx.sdsc.edu/docs/html/pages/usrgu068.htm#Header_440

gridData/OpenDX.py
```diff
@@ -182,18 +192,118 @@ def write(self,file):
...
+    # (round-tripping is not guaranteed to produce identical types)
+    np_types = {
+        "uint8": "byte",         # DX "unsigned byte" equivalent
+        "int8": "signed byte",
+        "uint16": "unsigned short",
+        "int16": "short",         # DX "signed short" equivalent
+        "uint32": "unsigned int",
+        "int32": "int",           # DX "signed int"   equivalent
+        "uint64": "unsigned int", # not explicit in DX, for compatibility
+        "int64": "int",           # not explicit in DX, for compatibility
```
kain88-de: are they stored in binary or plain values? Do we have enough spaces to write the maximal/minimal values of this type into the dx file?

orbeckst:
I'm on holidays and won't do much coding over the next few days. If you want to finish the PR then that's super-cool, otherwise I see if I can get to it after Xmas. 

- yes , one int needs to be removed, the last one as you said 

- DX stores ascii text and I just write with str()... maybe there are better ways. 

--
Oliver Beckstein
email: orbeckst@gmail.com

kain88-de:
Thanks @orbeckst

orbeckst:
I need to add docs (to show how things write files for pymol...) and I think I want to follow your earlier suggestion and not use type as a kwarg but instead use atype for "array type".

orbeckst:
rebased, cleaned up history... ready to merge.

(Decided to leave kwarg `type` but added extensive docs on #35 which can also be linked in PR https://github.com/MDAnalysis/mdanalysis/pull/1746)

Issue #47, opened 2018-05-04
create appveyor config


orbeckst:
In order to support native MDAnalysis windows builds (see ge PR https://github.com/MDAnalysis/mdanalysis/pull/1880) we also need a gridDataFormats native windows package.

The appveyor https://ci.appveyor.com/project/orbeckst/griddataformats project is enabled but currently fails.

orbeckst:
@tylerjereddy would you be able to help with getting GridDataFormats running on AppVeyor?

orbeckst:
Actually, as a pure Python package, pip install on windows should be fine. Still needs to be tested, though.

@tylerjereddy a quick look from you would be highly appreciated.

PR #52, opened 2018-07-29
Updated parsing for DX files

tluchko:
While using GridDataFormats, I ran into a couple of issues.

1. The library couldn't read in some of the files it produced.  I tracked this down to lines that look like
```
0.0     -1e-06  0.0     
```
In particular, the parser was splitting `-1e-6` into `-1` and `e-6`.  I modified the `FLOAT` regular expression but this required also modifying parts of the code that expected `INTEGER` as all `int`s  are also `float`s.

2. Parsing large files was pretty slow.  On my system, it would take 45 s to read a 74 MB file. I changed just the loop where the data is read to avoid invoking the parser and doing the type casting.  This got the total read time down to about 7 s.  If you are willing to assume that the data has three entries per line (except the last) you can get this down to about 4 s, of which 2 s is spent reading in the data.  My commit has the slower code that doesn't make this assumption and I've commented out the faster version.

For comparison, just reading the file into a list of strings takes about 1.5 s using the code
```
s.extend([line.split() for line in fh])
```
where `s` is an empty list and `fh` is an open file handle.

I wasn't sure how to run your test suite but I've used these changes on many of my own files and they work fine.

Hopefully you find this useful,

Tyler

### DIFF
--- a/gridData/OpenDX.py
+++ b/gridData/OpenDX.py
@@ -775,9 +775,9 @@ class DXParser(object):
             try:
                 while True:
                     # raises exception if not an int
-                    self.__peek().value('int')
+                    self.__peek().value('INTEGER')
                     tok = self.__consume()
-                    shape.append(tok.value('int'))
+                    shape.append(tok.value('INTEGER'))
             except (DXParserNoTokens,ValueError):
                 pass
             if len(shape) == 0:
@@ -830,9 +830,9 @@ class DXParser(object):
             try:
                 while True:
                     # raises exception if not an int
-                    self.__peek().value('int')
+                    self.__peek().value('INTEGER')
                     tok = self.__consume()
-                    shape.append(tok.value('int'))
+                    shape.append(tok.value('INTEGER'))
             except (DXParserNoTokens,ValueError):
                 pass
             if len(shape) == 0:
@@ -867,14 +867,14 @@ class DXParser(object):
         elif tok.equals('rank'):
             tok = self.__consume()
             try:
-                self.currentobject['rank'] = tok.value('int')
+                self.currentobject['rank'] = tok.value('INTEGER')
             except ValueError:
                 raise DXParseError('array: rank was "%s", not an integer.'%\
                                    tok.text)
         elif tok.equals('items'):
             tok = self.__consume()
             try:
-                self.currentobject['size'] = tok.value('int')
+                self.currentobject['size'] = tok.value('INTEGER')
             except ValueError:
                 raise DXParseError('array: items was "%s", not an integer.'%\
                                    tok.text)
@@ -889,9 +889,19 @@ class DXParser(object):
             if not self.currentobject['size']:
                 raise DXParseError("array: missing number of items")
             # This is the slow part.  Once we get here, we are just
-            # reading in a long list of floats but using the parser is expensive.
-            self.currentobject['array'] = [self.__consume().value('REAL') \
-                                           for i in range(self.currentobject['size'])]
+            # reading in a long list of numbers.  Conversion to floats
+            # will be done later when the numpy array is created.
+
+            # Don't assume anything about whitespace or the number of elements per row
+            self.currentobject['array'] = []
+            while len(self.currentobject['array']) <self.currentobject['size']:
+                 self.currentobject['array'].extend(self.dxfile.readline().strip().split())
+
+            # If you assume that there are three elements per row
+            # (except the last) the following version works and is a little faster.
+            # for i in range(int(numpy.ceil(self.currentobject['size']/3))):
+            #     self.currentobject['array'].append(self.dxfile.readline())
+            # self.currentobject['array'] = ' '.join(self.currentobject['array']).split()
         elif tok.equals('attribute'):
             # not used at the moment
             attribute = self.__consume().value()
--- a/gridData/OpenDX.py
+++ b/gridData/OpenDX.py
@@ -605,13 +605,16 @@ class DXParser(object):
     """
 
     # the regexes must match with the categories defined in the Token class
+    # REAL regular expression will catch both integers and floats.
+    # Taken from
+    # https://docs.python.org/3/library/re.html#simulating-scanf
     dx_regex = re.compile(r"""
     (?P<COMMENT>\#.*$)            # comment (until end of line)
     |(?P<WORD>(object|class|counts|origin|delta|type|counts|rank|items|data))
     |"(?P<QUOTEDSTRING>[^\"]*)"   # string in double quotes  (quotes removed)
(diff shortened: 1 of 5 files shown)


gridData/OpenDX.py
```diff
@@ -885,8 +888,20 @@ def __array(self):
                             'array: Only the "data follows header" format is supported.')
             if not self.currentobject['size']:
                 raise DXParseError("array: missing number of items")
-            self.currentobject['array'] = [self.__consume().value('REAL') \
-                                           for i in range(self.currentobject['size'])]
+            # This is the slow part.  Once we get here, we are just
```
orbeckst: The performance improvement here looks very good.

gridData/OpenDX.py
```diff
@@ -775,10 +773,12 @@ def __gridpositions(self):
         if tok.equals('counts'):
             shape = []
             try:
-                while self.__peek().iscode('INTEGER'):
+                while True:
```
orbeckst: using `try/except` instead of `if` looks good to me

gridData/OpenDX.py
```diff
@@ -775,10 +773,12 @@ def __gridpositions(self):
...
+                    # raises exception if not an int
+                    self.__peek().value('INTEGER')
                     tok = self.__consume()
-                    shape.append(tok.value())
-            except DXParserNoTokens:
+                    shape.append(tok.value('INTEGER'))
+            except (DXParserNoTokens,ValueError):
```
orbeckst: PEP8: space after comma

gridData/OpenDX.py
```diff
@@ -829,10 +828,12 @@ def __gridconnections(self):
...
```
orbeckst: PEP8: space after comma

gridData/OpenDX.py
```diff
@@ -885,8 +888,20 @@ def __array(self):
                             'array: Only the "data follows header" format is supported.')
             if not self.currentobject['size']:
                 raise DXParseError("array: missing number of items")
-            self.currentobject['array'] = [self.__consume().value('REAL') \
-                                           for i in range(self.currentobject['size'])]
+            # This is the slow part.  Once we get here, we are just
+            # reading in a long list of numbers.  Conversion to floats
+            # will be done later when the numpy array is created.
+
+            # Don't assume anything about whitespace or the number of elements per row
```
orbeckst: Yes – keep it more general. Thanks for the comments in the code.

gridData/OpenDX.py
```diff
@@ -611,10 +611,8 @@ class DXParser(object):
     |"(?P<QUOTEDSTRING>[^\"]*)"   # string in double quotes  (quotes removed)
     |(?P<WHITESPACE>\s+)          # white space
     |(?P<REAL>[-+]?               # true real number (decimal point or
-          (\d+\.\d*([eE][-+]\d+)?)  # scientific notation)
-          |(\d*\.\d+([eE][-+]\d+)?)
-          |(\d[eE][-+]\d+))
-    |(?P<INTEGER>[-+]?\d+)       # integer
```
orbeckst: I am not thrilled throwing out the INTEGER pattern. Could you add a test case with an example that fails with the original code, something that contains
```
0.0     -1e-06  0.0     
```
Maybe just update [gridData/tests/test.dx](https://github.com/MDAnalysis/GridDataFormats/blob/master/gridData/tests/test.dx) with some offending lines or add a new minimal tests file.

Basically, we want to have a test that shows that the bug has been squashed.

orbeckst: It should be possible to fix the regular expressions... I probably made a subtle mistake when I wrote them a long time ago.

gridData/OpenDX.py
tluchko: Fixed in `64fc3ae`

gridData/OpenDX.py
tluchko: Fixed in `64fc3ae`

gridData/OpenDX.py
tluchko: Sorry, do you want me to delete the alternate version or leave it as is?

gridData/OpenDX.py
tluchko: I modified `test.dx` in `0fabaa9`and `3598ae5`.  I don't know what framework you are using for your test suite or how to run the tests.  I am really only familiar with the `unittest` module. I can confirm that this test file now loads correctly but would fail at multiple points in the previous version.

I tried to keep the `INTEGER` pattern but anything that is an integer is captured by `FLOAT` since regular expressions are greedy.  I only deleted it because it didn't do anything after I updated the `FLOAT` expression. It might be possible to get `INTEGER` to work by looking ahead or back but I didn't investigate this. BTW, I just took the `FLOAT` regex from the [`re`](https://docs.python.org/3/library/re.html#simulating-scanf) documentation.

tluchko:
I also updated AUTHOR and CHANGELOG.  Following previous entries, I added new version number in CHANGELOG but I did not try to update this anywhere else.

orbeckst:
@tluchko the test [gridData.tests.test_dx.test_read_dx](https://travis-ci.org/MDAnalysis/GridDataFormats/jobs/414726557#L838) fails. Do you know why? Does it indicate that the bug is not fixed – or do you have to update the test so that it reflects the expected result?

gridData/OpenDX.py
orbeckst: Ok, thanks for the explanation. Then let's go with your modifications and leave out INTEGER.

Can you add a comment to the regexes that references the source as the re module with the link that you gave and that states that FLOAT will also catch integers?

orbeckst:
In the tests at https://github.com/MDAnalysis/GridDataFormats/blob/030628dfd458bfc4e1936f4e807a3e6272994e44/gridData/tests/test_dx.py#L15 you need to replace the `np.ones(POINTS)` with something like
```python
ref = np.ones(POINTS)
ref[4] = 1e-6
ref[5] =  1e+6
assert_equal(g.grid.flat, ref)
```

tluchko:
Is there a way to run the test suite locally?  Trying to guess and test by pushing is not working for me.

gridData/OpenDX.py
tluchko: Done.

orbeckst:
P.S.: Only the Travis tests have to pass, the AppVeyor Windows builds are not properly configured at the moment (see #47).

gridData/OpenDX.py
orbeckst: Leave it.

orbeckst:
@tluchko thank you for your first contribution!

PR #54, opened 2018-08-15
CCP4 byteorder detection

orbeckst:
- should help with debugging #50
- added test that checks that the test ccp4 file byteorder is properly detected


orbeckst:
This does not solve #50 but at least the better diagnostics and additional tests are now in master

PR #58, opened 2018-11-12
Work around more DX parsing issues by other apps

giacomofiorin:
The code below from the GridForcesGrid.C file in NAMD:
```
fscanf(poten_fp, "object %*d class array type double rank 0 "
                 "items %*d data follows\n");
```
will only work if the quote character (introduced in 115b05af to work around a similar problem in the PyMol parser - see #35) is omitted.  This is done via the optional `typequote` keyword, to accompany `type`.

The other commit has to do with another inconsistency in PyMol.  Its DX parser fails to parse tiny numbers (i.e. with an exponent lower than -100) despite using a `float` for parsing and the `%f` format in the function:
```
static int ObjectMapACNTStrToMap(ObjectMap * I, char *ACNTStr, int bytes, int state,
                                 int quiet);
```
The most recent versions of PyMol now support `type float`:
https://github.com/schrodinger/pymol-open-source/commit/4bbe42513ceb44ae84d51526e62b621e33e7df53
which is a reasonable decision, given that it is actually not parsing `double`s correctly :-)


giacomofiorin:
The amended commit now handles also test cases where grids of integer values are being used.

Note that the AppVeyor test fails due to Visual Studio configuration issues (probably unrelated to this PR).

giacomofiorin:
Hello, a brief reminder that the Travis test passes, whereas the AppVeyor test fails due to causes unrelated to this PR:
```
The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.
```

orbeckst:
@giacomofiorin sorry for the long silence – I haven't had a look at GridDataFormats in a while. In general, ping me with @orbeckst .

I'll have a look at your contribution, thank you!

gridData/core.py
```diff
@@ -437,12 +440,19 @@ def export(self, filename, file_format=None, type=None):
...
-
         .. _`#35`: https://github.com/MDAnalysis/GridDataFormats/issues/35
 
+        typequote : str (optional)
+            For DX, set the character used to quote the type string;
+            by default this is a double-quote character, '"'.
+            Custom parsers like the one from NAMD-GridForces (backend for MDFF)
+            expect no quotes, and typequote='' may be used to appease them.
+
+            .. versionadded:: after 0.4.0
```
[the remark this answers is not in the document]
orbeckst: With semantic versioning, we have to put this _new feature_ into the next minor release.

giacomofiorin:
@orbeckst Thanks. Both new features are used within `test_dx.py`, so that should be sufficient.

orbeckst:
Ok, then add yourself to AUTHORS and add an entry to CHANGELOG. I don't have an updated CHANGELOG yet so use the current one and just add your entry under 0.4.1 but I will put whatever you write in the correct spot before I merge.

giacomofiorin:
I did that in the merge commit, except I put 0.5.0 as suggested. Feel free to edit the description as needed.

gridData/core.py
```diff
@@ -426,10 +426,13 @@ def export(self, filename, file_format=None, type=None):
...
+
         filename : str
             name of the output file
+
         file_format : {'dx', 'pickle', None} (optional)
             output file format, the default is "dx"
+
         type : str (optional)
             for DX, set the output DX array type, e.g., "double" or
             "float"; note that PyMOL only understands "double" (see
```
orbeckst: In the light of your comment 
> The most recent versions of PyMol now support type float: schrodinger/pymol-open-source@4bbe425
which is a reasonable decision, given that it is actually not parsing doubles correctly :-)

is this comment in the docs still relevant?

gridData/core.py
```diff
@@ -439,12 +442,19 @@ def export(self, filename, file_format=None, type=None):
...
-
         .. _`#35`: https://github.com/MDAnalysis/GridDataFormats/issues/35
 
+        typequote : str (optional)
+            For DX, set the character used to quote the type string;
+            by default this is a double-quote character, '"'.
+            Custom parsers like the one from NAMD-GridForces (backend for MDFF)
+            expect no quotes, and typequote='' may be used to appease them.
+
+            .. versionadded:: after 0.5.0
```
orbeckst: Please change to
```reST
.. versionadded:: 0.5.0
```

gridData/core.py
giacomofiorin: It may still be relevant if the numbers in the DX map are outside the allowable range for a 32-bit float, regardless of which type is stated in the DX header.  You may want to actually *discourage* using `type=double` because Pymol doesn't really support it.

However, this is really a Pymol issue, and in all practical scenarios it is avoided now with the new format.  So I've just removed the comment for simplicity.

orbeckst:
@giacomofiorin thank you for your contribution. Sorry it took such a long time. Next time ping me with '@orbeckst' to get my attention.

PR #72, opened 2019-11-07
fix #71 init of core.Grid metadata

### LINKED ISSUE #71: core.Grid metadata badly initialized
instance variable metadata is shared between instances of Grid()

ReneHamburger1993:
fix for #71  prevent sharing of mutable dict between instances of core.Grid() class


gridData/core.py
```diff
@@ -131,7 +131,10 @@ def __init__(self, grid=None, edges=None, origin=None, delta=None,
             'PYTHON': self._load_python,  # compatibility
         }
 
-        self.metadata = metadata
+        if metadata is not None:
```
orbeckst: If you want to you can also write it as
```python
self.metadata = metadata if metadata is not None else {}
```

gridData/core.py
```diff
@@ -131,7 +131,10 @@ def __init__(self, grid=None, edges=None, origin=None, delta=None,
...
+            self.metadata = metadata
+        else:
+            self.metadata = {} # prevent sharing of mutable dict between instances
```
orbeckst: No need for the comment ;-)

gridData/core.py
ReneHamburger1993: done

orbeckst:
Thank you @ReneHamburger1993 !

PR #84, opened 2020-08-05
Make interpolation outside of known map 0.

jvermaas:
There are instances where "nearest" interpolation leads to some really wild behavior. For instance, I'm trying to add maps together, and the maps themselves are not guaranteed to have zeros along their edges. So when I try summing them together using their interpolated values into a map that is physically larger, extruded volumes of nonzero density stretch out to the edge of the new map. I am not sure under what scenario "nearest" or "wrap" would be correct. For my case, where I am adding together density maps, a "constant" interpolation is correct.

For instance, 

```python
#It is important here that the contents of "somefile" extend to the edge of the grid.
g1 = Grid("somefile")
g2 = Grid("somefile")
g2.origin += 5 #Shift the origin for demonstration purposes to fill the case where two grids have unequal centers.

#We want newG to be g1 + g2. Since the origin is shifted, this will be larger than the original contents of "somefile"
newshape = np.ceil(g2.origin + (g1.grid.shape * g1.delta)  - g1.origin) / g1.delta).astype('int')
neworigin = g1.origin
newG = Grid(grid=np.zeros(newshape, dtype=np.float), origin=neworigin, delta = g1.delta)
newG += g1.resample(newG)
newG += g2.resample(newG)
#At this stage, the "core" density will be right, but you'll see extrusions where the values pulled from the "nearest" interpolation is nonzero.
```

Now if there is some reason why "nearest" should be preferred, please let this behavior be modified through a passed parameter to `resample`.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -28,7 +28,7 @@ The rules for this file:
   * fix multiple __init__ calls (#73)
   * interpolation behavior outside of the grid changed to default to a
     constant rather than the nearest value (#84)
-  * corrected resampling behavior to not draw on values outside of the grid
+  * corrected resampling behavior to not draw on values outside of the grid (#84)
 
 
 05/16/2019 giacomofiorin, orbeckst
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,7 +13,7 @@ The rules for this file:
   * accompany each entry with github issue/PR number (Issue #xyz)
 
 ------------------------------------------------------------------------------
-??/??/2019 eloyfelix, renehamburger1993, lilyminium
+??/??/2019 eloyfelix, renehamburger1993, lilyminium, jvermaas
 
   * 0.6.0
 
@@ -26,6 +26,8 @@ The rules for this file:
 
   * fix initialization of mutable instance variable of Grid class (metadata dict) (#71)
   * fix multiple __init__ calls (#73)
+  * interpolation behavior outside of the grid changed to default to a
+   constant rather than the nearest value (#84)
 
 
 05/16/2019 giacomofiorin, orbeckst
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -27,7 +27,8 @@ The rules for this file:
   * fix initialization of mutable instance variable of Grid class (metadata dict) (#71)
   * fix multiple __init__ calls (#73)
   * interpolation behavior outside of the grid changed to default to a
-   constant rather than the nearest value (#84)
+    constant rather than the nearest value (#84)
+  * corrected resampling behavior to not draw on values outside of the grid
 
 
 05/16/2019 giacomofiorin, orbeckst
--- a/gridData/core.py
+++ b/gridData/core.py
@@ -267,8 +267,9 @@ class Grid(object):
         if float(factor) <= 0:
             raise ValueError("Factor must be positive")
         # Determine current spacing
-        spacing = (numpy.array(self._max_edges(
-        )) - numpy.array(self._min_edges())) / (-1 + numpy.array(self._len_edges()))
+        spacing = (numpy.array(self._max_edges()) \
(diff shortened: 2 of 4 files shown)


jvermaas:
Quick note: there isn't actually a test currently that looks outside the boundaries of the old grid when resampling. Do you have a specific test case in mind that you'd like to use? Otherwise, you wouldn't see the difference between "nearest" and "constant" interpolation.

orbeckst:
I just noticed that a test related to interpolation was failing, probably because it now produces different output. If you could have a look (e.g. https://travis-ci.org/github/MDAnalysis/GridDataFormats/jobs/715322838) and adjust the test accordingly.

jvermaas:
Ahh, ok. I now see why "nearest" or "wrap" was considered ideal. It is a bug in how `resample_factor` recalculates the edges. I'll get it sorted.

jvermaas:
@orbeckst, I think I've got it sorted now. Are the codecov checks complaining because there isn't a check currently to make sure we do indeed throw a value error when the scaling factor is negative?

gridData/core.py
```diff
@@ -257,12 +257,32 @@ def resample_factor(self, factor):
...
+           grid edges for every resampling.
+
         """
-        # new number of edges N' = (N-1)*f + 1
-        newlengths = [(N - 1) * float(factor) + 1 for N in self._len_edges()]
-        edges = [numpy.linspace(start, stop, num=int(N), endpoint=True)
-                 for (start, stop, N) in
-                 zip(self._min_edges(), self._max_edges(), newlengths)]
+        if float(factor) <= 0:
+            ValueError("Factor must be positive")
```
orbeckst, line 268: Exception needs to be raised:
```suggestion
            raise ValueError("Factor must be positive")
```

gridData/core.py
```diff
@@ -257,12 +257,32 @@ def resample_factor(self, factor):
...
+        # Determine current spacing
+        spacing = (numpy.array(self._max_edges(
+        )) - numpy.array(self._min_edges())) / (-1 + numpy.array(self._len_edges()))
```
orbeckst, line 271: Formatting is a bit odd, maybe
```suggestion
        spacing = (numpy.array(self._max_edges()) \
                   - numpy.array(self._min_edges())) / \
                  (-1 + numpy.array(self._len_edges())
```

orbeckst: ... sometimes just ignoring PEP8 is also acceptable ;-)

gridData/tests/test_grid.py
```diff
@@ -166,10 +166,14 @@ def test_resample_factor(self, data):
 
         g = data['grid'].resample_factor(2)
         assert_array_equal(g.delta, np.ones(3) * .5)
-        assert_array_equal(g.grid.shape, np.ones(3) * 6)
-        # check that the edges are the same
-        assert_array_almost_equal(g.grid[::5, ::5, ::5],
-                                  data['grid'].grid[::2, ::2, ::2])
+        # zooming in by a factor of 2. Each subinteral is
```
orbeckst, line 169:
```suggestion
        # zooming in by a factor of 2. Each subinterval is
```

CHANGELOG
```diff
@@ -26,6 +26,9 @@ The rules for this file:
 
   * fix initialization of mutable instance variable of Grid class (metadata dict) (#71)
   * fix multiple __init__ calls (#73)
+  * interpolation behavior outside of the grid changed to default to a
+    constant rather than the nearest value (#84)
+  * corrected resampling behavior to not draw on values outside of the grid
```
orbeckst, line 31:
```suggestion
  * corrected resampling behavior to not draw on values outside of the grid (#84)
```

orbeckst:
Many thanks, @jvermaas !

PR #90, opened 2021-09-14
[MAINT] Add gh actions CI

### LINKED ISSUE #86: MAINT: Switch from Travis to Github actions for CI
Follow up from: https://github.com/MDAnalysis/mdanalysis/issues/3036

As our Travis allocation is organisation wide, it might be best to switch over to Github actions where possible.

IAlibay:
Fixes #86

Changes:
  - Added GH actions CI
  - Replaced appveyor with gh actions windows runner (it's slightly different [we no longer test a pip only installation world], but this might just be cleaner?)
  - Added action for docs deployment
  - Added new badges, including NF


IAlibay:
appveyor ci setup is just broken, will try migrating to gh actions.

IAlibay:
If you agree to just removing appveyor @orbeckst we can also close #85 - we'll need to remove the hook for it in settings too.

I guess I can't test that docs actually works until this gets merged. We'll have to fix it live if I made a typo somewhere.

orbeckst (approved):
looks good — if we have to fix it live then we'll just do that, no big deal here

thanks a lot!

orbeckst:
This repo does not have any required checks it seems... probably something to update when the new CI is working.

Issue #91, opened 2021-09-14
Docs deployment broken


IAlibay:
No code has ever worked first time: https://github.com/MDAnalysis/GridDataFormats/runs/3604545377?check_suite_focus=true#step:8:185

PR #92, opened 2021-09-14
try to fix doc deployment

IAlibay:
I'll be honest I'm not 100% sure what's causing #91, I'm sure it's something super simple I'm not seeing.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -61,8 +61,8 @@ master_doc = 'index'
 # General information about the project.
 project = u'GridDataFormats'
 authors = (u'Oliver Beckstein, Jan Domanski, Jesse Johnson, Max Linke, Tyler Luchko, '
-           u'Dominik Mierzejewski, Giacomo Fiorin, Lily Wang')
-copyright = u'2007-2020, ' + authors
+           u'Dominik Mierzejewski, Giacomo Fiorin, Lily Wang, Josh Vermaas, Irfan Alibay')
+copyright = u'2007-2021, ' + authors
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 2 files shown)


IAlibay:
don't merge as-is, I've just enabled PR docs building to allow us to diagnose the issue.

orbeckst:
Isn't there a GH action for pages deployment that we could use instead?

IAlibay:
> Isn't there a GH action for pages deployment that we could use instead?

https://github.com/marketplace/actions/github-pages-action or https://github.com/marketplace/actions/deploy-to-github-pages seem to do the trick.

They seem widely enough used if you want to go with either of them.

orbeckst:
I'd be fine with either.

doc/source/conf.py
```diff
@@ -61,8 +61,8 @@
 # General information about the project.
 project = u'GridDataFormats'
 authors = (u'Oliver Beckstein, Jan Domanski, Jesse Johnson, Max Linke, Tyler Luchko, '
-           u'Dominik Mierzejewski, Giacomo Fiorin, Lily Wang')
-copyright = u'2007-2020, ' + authors
+           u'Dominik Mierzejewski, Giacomo Fiorin, Lily Wang, Josh Vermaas, Irfan Alibay')
```
IAlibay, line 64: Apparently the author list doesn't get auto-populated. Does this need an issue?
Also am I good to bump up the copyright to 2021?

orbeckst: bump it up

you can raise an issue for auto-populating, especially as we already have the code somewhere (and this code was put under a special license IIRC...)

orbeckst (approved):
cool, thanks!!!

IAlibay:
Let's see how this goes.. at least it builds now?

Issue #76, opened 2019-12-05
Wrong conversion from CCP4 to DX


nezix:
#### Importing a CCP4 map to convert it to a DX file outputs an inverted DX file.

```python
from gridData import Grid
g = Grid("1cbs.ccp4")
g.export("1cbs.dx")
```

When displaying the DX file in Pymol or VMD, the molecule does not fit in the iso-surface

Level 1 of the exported DX file in Pymol:
![image](https://user-images.githubusercontent.com/9949327/70249818-3d98db00-177d-11ea-917d-b659e9b211db.png)

Level 1 of the CCP4 file in Pymol:
![image](https://user-images.githubusercontent.com/9949327/70249920-6620d500-177d-11ea-82a4-e16fbf8413ee.png)

 Did I miss something ?

orbeckst:
That looks wrong... do you have an example file that you can add to the issue or link?

nezix:
You can find the ccp4 file here https://www.ebi.ac.uk/pdbe/coordinates/files/1cbs.ccp4

orbeckst:
I can confirm this bug.

Furthermore, the DX density appears to be stretched (see Chimera image, contours at 0.362)
- gray: [1cbs.ccp4](https://www.ebi.ac.uk/pdbe/coordinates/files/1cbs.ccp4)
- yellow: DX file from gridDataFormats
- licorice (barely visible) [1cbs.pdb](https://www.rcsb.org/structure/1CBS)

![1cbs_ccp4_dx](https://user-images.githubusercontent.com/237980/74881049-d3c6b100-5329-11ea-9d1d-e28ea02c792f.png)

orbeckst:
Do we know if the problem is with 
- CCP4 reading
- CCP4 interpretation (conversion to Grid)
- something else (probably not DX writing because that's been used widely)

orbeckst:
PR #100 does not fix this bug. In ChimeraX loading the densities gives
* CCP4 file: 81 x 70 x 96
* dx file: 70 x 81 x 96

Clearly there's something wrong in how gridData interprets the data in some(?) CCP4 files.

PR #100, opened 2021-12-30
use mrcfile library for MRC/CCP4 file parsing

### LINKED ISSUE #83: CCP4 file format is out of date
The most recent MRC/CCP4 file format is available at https://www.ccpem.ac.uk/mrc_format/mrc2014.php, which is a significant update from the format used in gridDataFormats.  CCP-EM maintains a Python library for reading, writing and validating MRC2014 files at https://github.com/ccpem/mrcfile.  A simple option is to use this library for reading *and* writing but this would add an external dependency to gridDataFormats.

I'd be happy to add in support for reading and writing MRC2014 files by using the `mrcfile` module or update the existing code (which would be read-only support).  Is there a preference?

orbeckst:
- fix #83
- use mrcfile https://github.com/ccpem/mrcfile (light-weight dependency) for file parsing
- new mrc module; functionality replaces CCP4 module
- keep CCP4 module for one release cycle
- add tests


orbeckst:
For reviewers: I intend to ignore the PEP8 formatting warnings... if you feel strongly about it then I suggest we run everything through black from here on out. I don't have the time to spare to manually prettify code that started out pretty ugly anyway.

orbeckst:
@MDAnalysis/coredevs if you want to have a quick look then that would be much appreciated. If you think it's good enough then you can approve it — I am mainly looking for a second pair of eyes to avoid really stupid mistakes.

If nobody has time I'll merged it by the end of the week latest in order to get a release out.

IAlibay:
> If nobody has time I'll merge it by the end of the week latest in order to get a release out.

I've penciled time in to review on Thursday if that's ok? (assuming no one else gets to review first)

tluchko:
@orbeckst, thank you for implementing this.  Doing so has been on my to-do list for a long time but never managed to get to it.

The code looks good to me but I have two questions about functionality.

1. Why are non-orthorhombic unit cells not supported?  It seems they are supported for the DX format, as I get no complaints when I load such a file.
2. Can we add support for writing MRC files?  I played around with this and I think it will be straightforward.  However, I don't know how to recover the unit cell angles for non-orthorhombic from a `Grid` object.

I can add writing MRC files for orthorhombic cells, if you like. Once I understand the issues for non-orthorhombic cells, I should be able to add read/write support for these as well.

gridData/mrc.py
```diff
@@ -0,0 +1,136 @@
...
+        self.filename = filename
+        if filename is not None:
+            self.read(filename)
+
+    def read(self, filename):
+        """Populate the instance from the MRC/CCP4 file *filename*."""
+        if filename is not None:
+            self.filename = filename
+        with mrcfile.open(filename) as mrc:
+            if not mrc.is_volume():                           #pragma: no cover
```
IAlibay, line 98: is calling validate worth it too? (not sure how much of an issue a non-valid mrc file is)

gridData/mrc.py
```diff
@@ -0,0 +1,136 @@
...
+                raise ValueError(
+                    "MRC file {} is not a volumetric density.".format(filename))
+            self.header = h = mrc.header.copy()
```
IAlibay, line 101: any difference between header and extended_header here?

gridData/mrc.py
IAlibay, line 98: Should there be an attempt to use `set_volume()` to change from an image stack to a volume?

gridData/mrc.py
```diff
@@ -0,0 +1,136 @@
...
+            # check for being orthorhombic
+            if not np.allclose([h.cellb.alpha, h.cellb.beta, h.cellb.gamma],
+                               [90, 90, 90]):
+                raise ValueError("Only orthorhombic unitcells are currently "
+                                 "supported, not "
+                                 "alpha={0}, beta={1}, gamma={2}".format(
+                                     h.cellb.alpha, h.cellb.beta, h.cellb.gamma))
+            # mrc.data[z, y, x] indexed: convert to x,y,z as used in GridDataFormats
```
IAlibay, line 109: Do we need to do further manipulations here?

```
Note that the MRC format allows the data axes to be swapped using the header’s mapc, mapr and maps fields. This library does not attempt to swap the axes and simply assigns the columns to X, rows to Y and sections to Z. (The data array is indexed in C style, so data values can be accessed using mrc.data[z][y][x].) In general, EM data is written using the default axes, but crystallographic data files might use swapped axes in certain space groups – if this might matter to you, you should check the mapc, mapr and maps fields after opening the file and consider transposing the data array if necessary.
```

gridData/tests/test_mrc.py
```diff
@@ -0,0 +1,97 @@
...
+    # label exists but contains more byte strings
+    #('label', ' Map from fft                                                                   '),
+    ('cella', np.rec.array((45.8, 45.8, 89.65),
+                        dtype=[('x', '<f4'), ('y', '<f4'), ('z', '<f4')])),
+    ('cellb', np.rec.array((90., 90., 90.),
+                           dtype=[('alpha', '<f4'), ('beta', '<f4'), ('gamma', '<f4')])),
+])
+def test_ccp4_read_header(ccp4data, name, value):
+    if type(value) is float:
+        assert_almost_equal(ccp4data.header[name], value, decimal=6)
```
IAlibay, line 89: I think the convention is now to use `assert_allclose` instead of `assert_almost_equal`.

I fee like there _should_ be a nicer way to do this comparison, but at the end of the day it works so there's no point trying to find the prettiest answer to this problem.

IAlibay (changes requested):
Everything but the one unnecessary import is a question.

gridData/mrc.py
tluchko: I agree.  If the file has `mapc, mapr, maps = 3, 2, 1`, then the data should already be in C-order and `np.transpose()` would be unnecessary.  `mapc, mapr, maps = 1, 2, 3` is Fortran-order.  It might be best to check for these two cases and disallow the rest. `np.transpose()` should be able to handle the other cases but it they are probably rare.

orbeckst:
Thanks for review and comments. This week I definitely won't have time to look at anything, sorry.

gridData/mrc.py
orbeckst: I don't want to deal with imagestacks — no experience and no idea how to test. I would leave this for someone who wants the feature.

orbeckst: You might want to read a file with an iffy header and just get the data. Not sure, TBH.

gridData/tests/datafiles/__init__.py
```diff
@@ -10,6 +10,8 @@
 # from http://www.ebi.ac.uk/pdbe/coordinates/files/1jzv.ccp4
 # (see issue #57)
 CCP4_1JZV = resource_filename(__name__, '1jzv.ccp4')
+# from https://github.com/ccpem/mrcfile/blob/master/tests/test_data/EMD-3001.map.bz2
+MRC_EMD3001 = resource_filename(__name__, 'EMD-3001.map.bz2')
```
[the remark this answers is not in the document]
orbeckst: BSD3 is fine inside GPL. We can add a note but it's more likely that the data does not fall under the license anyway.

gridData/mrc.py
orbeckst: don't know — experts??

gridData/mrc.py
orbeckst: Good point, this needs to be fixed then. Perhaps that fixes #76, too.

Code welcome!!!

orbeckst:
@tluchko can we try to get the simple reading code in and then add the enhancements, namely writing?

For the non-orthorhombic cells I can't quite remember why I restricted it to orthorhombic. There might be something in the Grid class that's not working with triclinic cells. I'd try to get this PR done with the limitation and then raise a new issue for triclinic cells.

gridData/mrc.py
tluchko: From https://www.ccpem.ac.uk/mrc_format/mrc2014.php, 

> The extended header is now used by different software to hold various additional metadata instead, as indicated by the EXTTYP tag. 

The `EXTTYP` flag takes seven different values. You could make a copy incase the user wants it.  Note that `self.header` does not seem to be accessible from `Grid` or, at least, it is not obvious to me how to access it.

tluchko (commented):
I suggested some code to deal with the different potential orderings of `mrc.data`.

gridData/mrc.py
```diff
@@ -0,0 +1,136 @@
...
+            self.array = np.transpose(mrc.data)
```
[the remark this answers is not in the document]
orbeckst: Thank you. I am adding your code, we'll still need a test.

Sorry to be slow, I have a ton of other things to attend to — beginning of semester is always stressful.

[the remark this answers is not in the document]
orbeckst: todo here
- [ ] test
- [ ] add @tluchko to CHANGELOG

gridData/tests/test_mrc.py
orbeckst: YEs, can be changed, of course. I copied & pasted what we had before.

gridData/mrc.py
orbeckst: Do you want to have access to `header` ?

gridData/core.py
```diff
@@ -123,7 +123,8 @@ def __init__(self, grid=None, edges=None, origin=None, delta=None,
             'PYTHON': self._export_python,  # compatibility
         }
         self._loaders = {
-            'CCP4': self._load_cpp4,
+            'CCP4': self._load_mrc,
```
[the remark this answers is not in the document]
orbeckst: I really only see this through the lens of the tests and if it doesn't break the tests then it didn't change anything. That's probably not true, though. There should definitely be a versionchanged entry to make sure users have a fighting chance to notice.

gridData/CCP4.py
```diff
@@ -160,6 +163,9 @@ class CCP4(object):
        * symmetry records
        * index ordering besides standard column-major and row-major
        * non-standard fields, such any in filed in future use block
+
+    .. deprecated:: 0.7.0
```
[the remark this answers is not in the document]
orbeckst: Need to check. We should have a RTD build in the PR – needs to be added to CI #104 .

gridData/mrc.py
tluchko: I don't have a personal-use case in mind.  But, if it isn't difficult, it could be beneficial in cases we haven't thought of yet. Not everything in the `header` ends up in a `Grid` object.

orbeckst:
The code with sorted mapc/r/s  transposed axes does _not_ fix #76 . In ChimeraX
* Opened **1cbs.ccp4** as 2, **grid size 81,70,96**, pixel 0.601,0.595,0.588, (= y, x, z)
* Opened **1cbs.dx** as 3, **grid size 70,96,81,** pixel 0.601,0.595,0.588 (=x, z, y)

1cbs.ccp4 :
* mapc, mapr, maps: 2, 1, 3.
* nx, ny, nz: 70, 81, 96
* mrc.data.shape = (96, 81, 70) (automatically reordered into z, x, y)

This looks as if mrcfile does indeed *always* reorder the data into z, x, y order.

gridData/CCP4.py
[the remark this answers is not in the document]
orbeckst: done

gridData/core.py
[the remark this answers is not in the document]
orbeckst: added versionchanged

gridData/mrc.py
```diff
@@ -0,0 +1,136 @@
...
+.. _`10.1107/S2059798317007859`: https://doi.org/10.1107/S2059798317007859
+
+Classes
+-------
+
+"""
+from __future__ import absolute_import
+from six.moves import range
+
+import warnings
```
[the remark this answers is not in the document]
orbeckst: removed

gridData/mrc.py
orbeckst: No implementing for now because I don't understand the implication/standard use.

gridData/mrc.py
orbeckst: I am adding the header as undocumented `Grid._mrc_header`. If it is used then we can think of a better way to keep format-specific data.

gridData/tests/test_mrc.py
orbeckst: kept it,,, need sleep

orbeckst:
@tluchko and @IAlibay I addressed most of your comments and after a long time thinking about what mrcfile does and what the mapc/r/s parameter mean I managed to get the conversion of MRC data to Grid to work (and fixed the super-irritating #76 along the way). This should be good for a final review.

If you can find a nicer way to write the double-transposition
```python
            axes_order = np.hstack([h.mapc, h.mapr, h.maps])
            axes_c_order = np.argsort(axes_order)
            self.array = np.transpose(np.transpose(mrc.data), axes=axes_c_order)
```
then let me know. However, just reversing `axes_c_order` does not work as far as I can tell... there must be a way to permute the `axes_c_order` directly but I am too tired to figure it out. (The first transpose is always (210) (rearrange zyx to xyz), the second one depends on the mapc/r/s values and is in the test (102) (x is really y, y is really x, and z is z). The end result needs to be (201) ((zyx) -> (yxz)).

gridData/mrc.py
orbeckst: Ultimately, this and calculating the origin correctly did fix #76

orbeckst:
Honestly, I don't know why reversing `axes_order_c` did not work for me because (102) reversed is (201) and that's also what you get when applying (210) to (102). But 
```python
self.array = np.transpose(mrc.data, axes=axes_c_order[::-1])
```
gives the wrong shape.

gridData/mrc.py
```diff
@@ -107,14 +105,24 @@ def read(self, filename):
...
+            axes_order = np.hstack([h.mapc, h.mapr, h.maps])
+            axes_c_order = np.argsort(axes_order)
+            self.array = np.transpose(np.transpose(mrc.data), axes=axes_c_order)
+            self.delta = np.diag(
+                np.array([mrc.voxel_size.x, mrc.voxel_size.y, mrc.voxel_size.z])[axes_c_order])
+            # the grid is shifted to the CCP4 origin by offset
+            # (assume orthorhombic)
+            offsets = np.hstack([h.nxstart, h.nystart, h.nzstart])[axes_c_order] * np.diag(self.delta)
+            # GridData origin is centre of cell at x=col=0, y=row=0 z=seg=0
+            self.origin = np.hstack([h.origin.x, h.origin.y, h.origin.z])[axes_c_order] + offsets
```
tluchko, line 125: The code below works with only one transpose.  Because `mrcfile` transposes the data, we need to reverse the order of `axes_order` before sorting. Note that `voxelsize`  and `origin`  refer to the x, y, and z real space axes and should not be reordered.  However, `nxstart`, `nystart` and `nzstart` should be reordered using `axes_c_order`. I'm not sure why, but I had to do this to match what ChimeraX and VMD are reporting for the example file in #76.
```suggestion
            transpose_order = np.argsort(axes_order[::-1])
            self.array = np.transpose(mrc.data, axes=transpose_order)
            self.delta = np.diag(np.array([mrc.voxel_size.x, mrc.voxel_size.y, mrc.voxel_size.z]))
            # the grid is shifted to the MRC origin by offset
            # (assume orthorhombic)
            offsets = np.hstack([h.nxstart, h.nystart, h.nzstart])[axes_c_order] * np.diag(self.delta)
            # GridData origin is centre of cell at x=col=0, y=row=0 z=seg=0
            self.origin = np.hstack([h.origin.x, h.origin.y, h.origin.z]) + offsets
```

tluchko (changes requested):
I suggested some revised code to avoid the double transpose of the data and match the delta and origin reported for test files in ChimeraX and VMD.

gridData/mrc.py
orbeckst: Thanks, I am going with your suggestion then. I am not an expert on the MRC format and from the description it wasn't clear to me what corresponds to which direction. 

I didn't include the data from #76 as a new test because it looked as if the included test file 1jzv.ccp4 for CCP4 would be sufficient (it also contains reordered axes). However, if you think that there's something in 1cbs.ccp4 that would make for a good test case then we can include it (especially if 1CBS showed you problems that were not apparent with 1jzv.ccp4).

orbeckst:
I rewrote the history so that this can be merged as is.

gridData/tests/test_mrc.py
```diff
@@ -0,0 +1,137 @@
+from __future__ import absolute_import, division
+
+import pytest
+
+import numpy as np
+from numpy.testing import (assert_almost_equal,
```
IAlibay, line 6:
```suggestion
from numpy.testing import (assert_allclose,
```

gridData/tests/test_mrc.py
```diff
@@ -0,0 +1,137 @@
...
+        b'                                                                                ',
+        b'                                                                                ',
+        b'                                                                                ',
+        b'                                                                                ',
+        b'                                                                                '],
+                    dtype='|S80'))
+])
+def test_ccp4_read_header(ccp4data, name, value):
+    if type(value) is float:
+        assert_almost_equal(ccp4data.header[name], value, decimal=6)
```
IAlibay, line 89:
```suggestion
        assert_allclose(ccp4data.header[name], value, rtol=1e-06)
```

gridData/tests/test_mrc.py
```diff
@@ -0,0 +1,137 @@
...
+        assert_equal(ccp4data.header[name], value)
+
+def test_axes_orientation(ccp4data):
+    # correctly interpret mapc, mapr, maps = 2, 1, 3
+    # for nx, ny, nz = 96, 76, 70.
+    # see also #76
+    assert_equal(ccp4data.shape, (76, 96, 70))
+
+def test_delta(ccp4data):
+    assert_almost_equal(ccp4data.delta, np.array(
```
IAlibay, line 100:
```suggestion
    assert_allclose(ccp4data.delta, np.array(
```

gridData/tests/test_mrc.py
```diff
@@ -0,0 +1,137 @@
...
+        [[0.5452381, 0.       , 0.       ],
+         [0.       , 0.5452381, 0.       ],
+         [0.       , 0.       , 0.5603125]], dtype=np.float32))
+
+def test_origin(ccp4data):
+    # shift with nxstart, nystart, nzstart and delta
+    #
+    # (visual comparison of CCP4 and DX file in ChimeraX at same
+    # level shows full agreement)
+    assert_almost_equal(ccp4data.origin, [-12.5404758,  -2.1809523,  57.151876 ])
```
IAlibay, line 110:
```suggestion
    assert_allclose(ccp4data.origin, [-12.5404758,  -2.1809523,  57.151876 ])
```

gridData/tests/test_mrc.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_shape(self, grid, ccp4data):
+        assert_equal(grid.grid.shape, ccp4data.shape)
+
+    def test_mrc_header(self, grid, ccp4data):
+        # undocumented MRC header in Grid
+        assert grid._mrc_header == ccp4data.header
+
+    def test_delta(self, grid, ccp4data):
+        assert_almost_equal(grid.delta, np.diag(ccp4data.delta))
```
IAlibay, line 131:
```suggestion
        assert_allclose(grid.delta, np.diag(ccp4data.delta))
```

gridData/tests/test_mrc.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_origin(self, grid, ccp4data):
+        assert_almost_equal(grid.origin, ccp4data.origin)
```
IAlibay, line 134:
```suggestion
        assert_allclose(grid.origin, ccp4data.origin)
```

gridData/tests/test_mrc.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_data(self, grid, ccp4data):
+        assert_almost_equal(grid.grid, ccp4data.array)
```
IAlibay, line 137:
```suggestion
        assert_allclose(grid.grid, ccp4data.array)
```

gridData/tests/test_mrc.py
```diff
@@ -0,0 +1,137 @@
...
+    _test_ccp4(g1)
+
+def test_ccp4_mrc(g2):
+    _test_ccp4(g2)
+
+def _test_ccp4(g):
+    POINTS = 192
+    assert_equal(g.grid.flat, np.arange(1, POINTS+1))
+    assert_equal(g.grid.size, POINTS)
+    assert_almost_equal(g.delta, [3./4, .5, 2./3])
```
IAlibay, line 34:
```suggestion
    assert_allclose(g.delta, [3./4, .5, 2./3])
```

IAlibay (approved):
I'll be honest here I don't really have enough time to fully understand what is being done with the MRC file transpose. From a quick look it seems to make sense and being no MRC expert I am happy to default to yours and @tluchko's expertise here.

I'm not sure what was meant by https://github.com/MDAnalysis/GridDataFormats/pull/100/files#r810475496 but I've added a couple of suggested changes for assert_almost_equal -> assert_allclose if it's still of interest to make the switch.

CHANGELOG
```diff
@@ -13,6 +13,25 @@ The rules for this file:
   * accompany each entry with github issue/PR number (Issue #xyz)
 
 ------------------------------------------------------------------------------
+??/??/2022 orbeckst, tluchko
```
orbeckst, line 16:
```suggestion
??/??/2022 orbeckst, tluchko, IAlibay
```

orbeckst:
Thank you @tluchko and @IAlibay !

orbeckst:
Following up on https://github.com/MDAnalysis/GridDataFormats/pull/100#issuecomment-1009487765, @tluchko please open issues for
* writing MRC files
* handling of triclinic boxes

Thanks!

gridData/mrc.py
tluchko: It looks like 1cbs.ccp4 doesn't add anything that isn't in 1jzv.ccp4 already. Both have the same use of `mapc`/`r`/`s`, `nstart` and `origin`. 

Looking back, `nx`, `ny`, `nz`, `nstart` and the data are all in reference to the grid, so the transpose/reorder specified by `mapc`/`r`/`s` must be applied.  OTOH, `mx`, `my`, `mz`, `cella`, `cellb` and `origin` refer to the real-space dimensions and do not require reordering. It would be nice to have additional test cases but I think these properties are handled correctly now.

tluchko:
Thank you @orbeckst!

gridData/mrc.py
orbeckst: Thanks for double checking, much appreciated!

PR #110, opened 2022-04-30
removed CCP4 module

### LINKED ISSUE #107: remove CCP4 module
The CCP4 module was replaced with mrc in 0.7.0 in PR #100 and deprecated ("to be removed in 0.8.0"). It now needs to be removed.

This will hopefully also "solve" #50 ...

orbeckst:
- fix #107 and #50
- removed CCP4.py and test_CCP4.py
- updated docs
- updated CHANGELOG


IAlibay:
Ah didn't see you had requested a review for this too, I'll get to it tomorrow UK time

orbeckst:
I don't think that a mrcfile release with the fixed https://github.com/ccpem/mrcfile/issues/35 is already out (the last one is 1.3.0 from Feb 2021 on https://pypi.org/project/mrcfile/) so we should not require a minimum version number (even though I think technically that would be the correct thing to do in order to claim having fixed #50 ).

orbeckst:
@IAlibay @richardjgowers I am looking for a rubberstamp here ;-)

IAlibay (approved):
Just the one tiny thing - approving to avoid blocking.

orbeckst:
Thanks Irfan!

PR #111, opened 2022-04-30
adopt NEP 29

### LINKED ISSUE #102: adopt NEP29
In order to make it easier on downstream dependencies (e.g., MDAnalysis/mdanalysis#3457) and to go along with #101 , remove support for legacy versions of Python (2.7, 3.5-3.6) by adopting [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html). 

- [ ] update CHANGELOG
- [ ] remove `six`
- [ ] remove unsupported test matrix entries

orbeckst:
- fix #102
- removed support for Python 2.7 and Python <3.8
  - removed six as dependency (setup.py and CI)
  - removed all `from __future__ import ...` statements
  - removed special Python division handling (and tests)
- update CHANGELOG
- set pip metadata to MATURE


orbeckst:
Also need to disable test requirement for dropped versions.

orbeckst:
Done

setup.py
```diff
@@ -45,7 +41,7 @@
       package_data={'gridData': ['tests/datafiles/*.dx', 'tests/datafiles/*.dx.gz',
                                  'tests/datafiles/*.ccp4', 'tests/datafiles/*.plt',
                                  'tests/datafiles/*.bz2']},
-      install_requires=['numpy>=1.0.3', 'six', 'scipy', 'mrcfile'],
+      install_requires=['numpy>=1.19', 'scipy', 'mrcfile'],
```
orbeckst, line 44: I bumped numpy because of NEP29 but it’s not really necessary. Could be reverted to better show what’s minimally needed.

IAlibay: My suggestion would be to follow NEP29 as closely as possible, if there ends up being issues further down the line then it can be relaxed?

.github/workflows/gh-ci.yaml
```diff
@@ -23,16 +23,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, ]
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9, "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
IAlibay, line 26: According to the schedule we are not py3.8+ https://numpy.org/neps/nep-0029-deprecation_policy.html
```suggestion
        python-version: ["3.8", "3.9", "3.10"]
```

IAlibay (changes requested):
Just the one thing and then I think it's good to go!

setup.py
orbeckst: Ok, thanksgiving

orbeckst:
Thanks @IAlibay i updated the metadata to show that we only support 3.8+.

IAlibay (approved):

Issue #112, opened 2022-05-24
Update setup to be NEP29 compliant


IAlibay:
Two things that should be done here to make sure that griddataformats is now py3.8+ only:

1) Drop the `universal` entry under `[bdist_wheel]` in `setup.cfg`. This should be for pure python builds that support both python 2 and 3 (https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#wheels)

From now on wheels should probably be built either manually or via actions (I'd be tempted in using griddataformats as a testing grounds for an actions based wheel deployment (so we can do the same with the core MDA repo) if that would be acceptable @orbeckst?)

2) Set a `python_requires` in `setup.py` to ensure that we don't install anything lower than py3.8 (https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires)

IAlibay:
As discussed here: https://github.com/conda-forge/griddataformats-feedstock/pull/11#issuecomment-1135251993 the current release of griddataformats is using a universal wheel with no limitations on the minimum python version, so if you pip installed griddataformats with an unsupported version of python you'd still pick it up. Doing these two fixes _should_ fix this long term.

orbeckst:
Ok, then we also have to make a 1.0.1 so that installation works properly. 

Thanks for the analysis and advice. I agree with everything you suggest. I’d get the 1.0.1 out sooner than later because otherwise we don’t have proper conda packages.

IAlibay:
I'll do the minimum changes now, and follow up with the wheel building action.

IAlibay:
Success! https://test.pypi.org/project/GridDataFormats/1.0.1b0/ and it works.

orbeckst:
awesomeness!

IAlibay:
@orbeckst I've just created a tag for 1.0.1, and it's created this upload on testpypi: https://test.pypi.org/project/GridDataFormats/1.0.1/

If you want to test it, I've updated the docs in the wiki page to reflect how to do so: https://github.com/MDAnalysis/GridDataFormats/wiki/Developer-Guide

Once you're happy, shall I leave it to you to cut a github release for 1.0.1? That _should_ automatically deploy the tarball and wheel to pypi (assuming the trigger works).

orbeckst:
Thanks for updating the dev docs.

I tested from testpypi and then created the 1.0.1 release, which auto-deployed to https://pypi.org/project/GridDataFormats/ 🚀 

[...]

PR #113, opened 2022-05-24
Strict python minimum version for install (Issue #112) and auto-deployment via actions

IAlibay:
Add `python_requires`, drops `universal` entry for wheel generation (note we probably should clean that out of the core repo's setup.cfg too...)

To do:
  - [ ] test that this works on testpypi (@orbeckst could you give me access on both pypi and testpypi? my user name is IAlibay on both)
  - [x] test that this works for wheel generation using build


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,57 @@
+name: Build and upload to PyPI
```
IAlibay, line 1: pypa has some really good docs on this it was a surprising experience: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

This will run a build runner every time we push/PR, and also whenever there's a release. If we push a new tag it will upload the build to testpypi for us to check before a release, if we pusblish a release it'll be automatically pushed to pypi.

What needs to be done here is:
1. Uploading the relevant github secret API tokens for testpypi and pypi.
2. Checking that the testpypi upload works (create a fake tag for an alpha version of 1.0.1 upload it and then delete the tag?)
3. Checking that the pypi upload works on publication of the next release.

orbeckst:
@IAlibay I invited you as owner on PyPi for gdf

orbeckst:
No, there's no testpypi.

IAlibay:
> Will you be able to check the new workflow once you have accepted ownership of gdf or do you need me to do something here?

Yeah I'll need to create a testpypi version but that should be fine

IAlibay:
Ok I've removed the requirement to review the build action because otherwise this will never end. Everything _seems_ to work right now, I just need to create the 1.0.1 release tag and see if testpypi likes it.

PR #114, opened 2022-05-24
bump python to 3.8 for docs action

IAlibay:
Not sure why that didn't fail in #113...

Anyways whilst I'm bothering you here @orbeckst - are there any docs that outline out griddataformats does a release deployment?

Do you just do a tag and then release / upload to pypi and that's it?

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -2,7 +2,13 @@ name: Build and upload to PyPI
 
 on:
   push:
+    branches:
+      - master
+    tags:
+      - *
   pull_request:
+    branches:
+      - master
   release:
     types:
       - published
@@ -21,7 +27,7 @@ defaults:
 jobs:
   build_wheels:
     environment: deploy
-    if: "github.repository == 'MDAnalysis/GridDataFormats'"
+    if: "github.repository == 'MDAnalysis/GridDataFormats'" && startsWith
     name: Build pure Python wheel
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


orbeckst:
@IAlibay here's the tag & release process https://github.com/MDAnalysis/GridDataFormats/wiki/Developer-Guide --- obviously, this will need to be changed when the new process is in place.

orbeckst (approved):
✔️

IAlibay:
ok thanks - I might have to iterate on the triggers for the deploy action, if you're ok with it I might have to self-merge a few PR until it's correct. I'll limit any commits to that one file.

Once the 1.0.1-beta tag is properly triggering an upload to testpypi I'll stop, confirm with you that it's all good and then we can do the 1.0.1 release?

orbeckst:
Yes, please go ahead!

orbeckst:
I saw in the deploy logs 
```
set build/lib/gridData/_version.py to '0+untagged.1.gd4c3aa5'
```
Shouldn't this be set to a proper tagged version, or at least build on the release-1.0.0 tag and look like 1.0.0+x.deadbeef ?

IAlibay:
> I saw in the deploy logs
> 
> ```
> set build/lib/gridData/_version.py to '0+untagged.1.gd4c3aa5'
> ```
> 
> Shouldn't this be set to a proper tagged version, or at least build on the release-1.0.0 tag and look like 1.0.0+x.deadbeef ?

Huh strange, I'm not sure why this is happening, will need to double check.

PR #115, opened 2022-05-24
Try to fix deployment tag version

IAlibay:
I suspect the issue is with the checkout fetch depth.


IAlibay:
"GridDataFormats-1.0.0+5.gec7027d" - that should fix it, now to see if I can trigger a release to testpypi with a tag

PR #120, opened 2022-08-01
updated pypi-publish action

### LINKED ISSUE #119: Outdated deploy action?
I had a look at the deployment to TestPyPI used here to build something similar on one of my repos. I got the following error in my own repo:
```
 You are using "pypa/gh-action-pypi-publish@master". The "master" branch of this project has been sunset and will not receive any updates, not even security bug fixes. Please, make sure to use a supported version. If you want to pin to v1 major version, use "pypa/gh-action-pypi-publish@release/v1". If you feel adventurous, you may opt to use use "pypa/gh-action-pypi-publish@unstable/v1" instead. A more general recommendation is to pin to exact tags or commit shas.
```

The deploy action seem to be outdated and it will likely fail at the next deployment. I think the deploy action needs to be updated, as it has been done with MDAnalysis which uses the following:
```
pypa/gh-action-pypi-publish@v1.5.0
```

orbeckst:
- fix #119
- use pypa/gh-action-pypi-publish@release/v1 as suggested in the deprecation note

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -52,13 +52,13 @@ jobs:
       - name: publish_testpypi
         # Upload to testpypi on every tag
         if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
-        uses: pypa/gh-action-pypi-publish@master
+        uses: pypa/gh-action-pypi-publish@release/v1
         with:
           password: ${{ secrets.TEST_PYPI_API_TOKEN }}
           repository_url: https://test.pypi.org/legacy/
 
       - name: publish_pypi
         if: github.event_name == 'release' && github.event.action == 'published'
-        uses: pypa/gh-action-pypi-publish@master
+        uses: pypa/gh-action-pypi-publish@release/v1
         with:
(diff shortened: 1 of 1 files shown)


orbeckst:
Note that @RMeli suggested a more specific pin to v1.5.0 as is used in MDA.  I went with the more generic v1 but would be happy to change, of course.

IAlibay (changes requested):
Whilst we are here, can you remove lines 4-6 please? We don't need the deploy action to run on push to master.

.github/workflows/deploy.yaml
```diff
@@ -2,8 +2,6 @@ name: Build and upload to PyPI
 
 on:
   push:
-    branches:
-      - master
```
orbeckst, line 6: I removed lines 5&6 instead of 4–6 — I assume the `push:` still has to be present, @IAlibay

IAlibay: Oops sorry yeah 5-6 (second time I mess up on this, accidentally told the same to @RMeli this morning 🤣 )

IAlibay (approved):
lgtm thanks!

PR #126, opened 2023-09-03
Switch deployment workflow to re-usable action.

IAlibay:
This switches the deployment to a re-usable action for a little bit more consistency across our ecosystem, and also some additional tests.


.github/workflows/deploy.yaml
```diff
@@ -8,55 +8,27 @@ on:
...
+      - name: testpypi_deploy
+        uses: MDAnalysis/pypi-deployment@main
         if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
-        uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-          repository_url: https://test.pypi.org/legacy/
+          token: ${{ secrets.TEST_PYPI_API_TOKEN }}
+          test_submission: true
+          package_name: 'gridData'
```
lilyminium, line 27: Isn't this gridDataFormats?

.github/workflows/deploy.yaml
```diff
@@ -8,55 +8,27 @@ on:
...
 
-      - name: publish_pypi
+      - name: pypi_deploy
+        uses: MDAnalysis/pypi-deployment@main
         if: github.event_name == 'release' && github.event.action == 'published'
-        uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          password: ${{ secrets.PYPI_API_TOKEN }}
+          token: ${{ secrets.PYPI_API_TOKEN }}
+          package_name: 'gridData'
```
lilyminium, line 34: Same question above -- this variable is used for `pip install`, right?

IAlibay: No it's used for the `pytest --pyargs` call, I _think_ this is correct.

IAlibay: ah wait I messed up somewhere

IAlibay: blergh what a pain - I'll have to change the action to account for this... funilly enough it didn't trip up mda_xdrlib at all.

IAlibay: Huh I guess I never fully realised pip doesn't care between `-` and `_` in package names.

lilyminium: Ah, I can't believe I keep spending the effort on typing underscores now 😂 better safe than sorry I guess

orbeckst (approved):
Let's try it.

PR #135, opened 2023-10-23
use MDA docs in README.rst

orbeckst:
- use our own docs in banner (use io shield)
- removed RTD-generated banner and link


lilyminium (approved):
LGTM -- thanks for the fix

PR #139, opened 2024-11-11
Bump to miniforge and py3.13

IAlibay:


orbeckst (approved):
I assume that the failures on 3.13 are because we do not have MDAnalysis at 3.13 yet?

Everything looks good in principle so I approve.

Thank you for the comprehensive update.

orbeckst:
@IAlibay please merge at will.

IAlibay:
> I assume that the failures on 3.13 are because we do not have MDAnalysis at 3.13 yet?

no.. it should be the other way around (griddataformats is a dependency of MDA) - I'll need to look at that more.

IAlibay:
I think I remember this bug being that conda isn't aliased to micromamba for non-linux cases. Hopefully this fixes it.

IAlibay:
Here we go, that does the trick - thanks for the review @orbeckst !

PR #149, opened 2026-01-13
Better volumetric test

### LINKED ISSUE #150: GridDataFormats refuses to load 3D MRC files with space group 0
MDAnalysis strictly follows the [CCP4](https://ftp.ebi.ac.uk/pub/databases/emdb/doc/Map-format/current/EMDB_map_format.pdf) and [MRC file specs](https://www.ccpem.ac.uk/mrc-format/mrc2014/) which state that files with a value of zero for the ISPG header field are stacks of 2D images, not single 3D volumes. However, many cryo-EM and volume visualization packages write a zero to this field because there is no "space group unknown" value. Because of this deviation from spec, most visualization packages open space group 0 files as volumes. For more from their point of view, see [this bug ticket for ChimeraX](https://www.rbvi.ucsf.edu/trac/ChimeraX/ticket/19713#comment:2).

Downstream users of GridDataFormats may be surprised that it refuses to open volumes that successfully open elsewhere. While GridDataFormat's behavior is formally correct, I think in practice it may be better to open any MRC file with a 3D data array as a 3D volume.

PlethoraChutney:
Hello! Thank you for MDAnalysis! This PR changes how MRC files are checked before being loaded as a 3D grid. Rather than use the `is_volume()` method from `mrcfile` we instead use a simpler shape check.

The reason for this is that `mrcfile` returns `False` if the ISPG header field of a 3D MRC file is zero. While this is correct according to the [spec](https://www.ccpem.ac.uk/mrc-format/mrc2014/), most cryoEM software packages save a 0 in this field for a variety of reasons. Because of this unwritten deviation from spec, most packages also read in files with a 0 space group as 3D Volumes anyway, but GridDataFormats does not.

Despite the fact that this is formally correct, I'm proposing that MDAnalysis should follow the behavior that users likely expect from their experience with other packages.

This PR fails three tests, but on my machine at least the current master branch also fails these same tests.

- gridData/tests/test_grid.py::TestGrid::test_load_wrong_fileformat[ccp4]
- gridData/tests/test_grid.py::TestGrid::test_load_wrong_fileformat[plt]
- gridData/tests/test_grid.py::TestGrid::test_load_wrong_fileformat[dx]


orbeckst:
@PlethoraChutney thank you for your contribution and the insights. Could you please raise an issue that describes the issue — basically your justification for the fix? Separating solution from issues is good practice when we're discussing changes that may go against specs. Issues are our permanent record for decisions.

We can then link the PR to the issue.

Once you have the issue up, we can look closer at your proposed solution. Thank you!

orbeckst (changes requested):
Blocked until we have discussed the approach.

BradyAJohnston:
It's related to this issue (https://github.com/BradyAJohnston/MolecularNodes/issues/993) raised in MN. Downloading directly from the EMDB it would load fine, but if we round-trip through ChimeraX then it would fail to load with `Grid()` because of the points specified above.

Worth some proper discussion though!

orbeckst:
Thanks for the context — ideally we collect all of this in an issue and then we have one place that documents the decision and the reasoning, as opposed to potentially mixing it with implementation details etc.

PlethoraChutney:
Apologies! I'm relatively new to actually contributing to others' repos. Created #150 for discussion

orbeckst:
Many thanks @PlethoraChutney , #150 is perfect. 

Different projects have different "cultures" on how they do development and when coming in you just don't know.

PlethoraChutney:
Tried to capture the discussion in #150 

- `Grid()` now takes a `**kwargs` parameter which is passed to all load functions. They all ignore it except MRC
- `MRC()` takes `load_mrc: None | bool`. By default (None), `mrcfile.is_volume()` is used. If a bool is provided, it forces the file to be a volume (True) or a 2D image stack (False)
- Added tests to cover this behavior

gridData/core.py
```diff
@@ -196,7 +196,7 @@ class Grid(object):
 
     def __init__(self, grid=None, edges=None, origin=None, delta=None,
                  metadata=None, interpolation_spline_order=3,
-                 file_format=None):
+                 file_format=None, **kwargs):
```
BradyAJohnston: I think I would prefer an actual argument rather than `**kwargs` as the latter can easily (and silently) absorb spelling mistakes `fiel_format='MRC` would be silently no longer passed on instead of raising an error otherwise. I defer to @orbeckst on this but I think just mentioning in the docs that it is only used for mrc files (and maybe a warning when used for non-compatible format?).

orbeckst: You're right that `**kwargs` can silently swallow typos. On the other hand, tracking individual modifier keywords for different parsers can get annoying, at least in principle. In MDAnalysis we've typically been using the approach of named keywords for arguments that universally apply and kwargs for modifiers.

GDF really does not see a flurry of development activity so the maintainability argument that I am raising is not particularly strong and I do see the value of being explicit and more robust for users. So then let's follow your idea and make `is_volume` a top-level kwarg of `Grid`, document it there and in MRC (with a proper versionadded), and propagate it explicitly.

BradyAJohnston: I feel strongly about it mostly because I am endlessly spending hours bug fixing when it was just me with simple spelling mistakes.

orbeckst: Per conversation: explicit `is_volume=None` kwarg.

Document it in the Grid docs with something like
```reST
is_volume : bool
    Indicate that the data to be loaded from a file is a density,
    even if the file may indicate differently. Currently only 
    applies for MRC files (see :class:`gridData.MRC.MRC`)

    The default ``None`` indicates that the data is handled
    according to the file format specifications.

    .. versionadded:: 1.1.0
```

gridData/core.py
```diff
@@ -238,7 +238,7 @@ def __init__(self, grid=None, edges=None, origin=None, delta=None,
                     filename = str(grid)
 
             if filename is not None:
-                self.load(filename, file_format=file_format)
+                self.load(filename, file_format=file_format, **kwargs)
```
orbeckst: Per conversation: explicit `is_volume=is_volume` kwarg.

gridData/core.py
```diff
@@ -545,32 +545,32 @@ def load(self, filename, file_format=None):
...
             saved = pickle.load(f)
         self._load(grid=saved['grid'],
                    edges=saved['edges'],
                    metadata=saved['metadata'])
 
-    def _load_mrc(self, filename):
+    def _load_mrc(self, filename, **kwargs):
         """Initializes Grid from a MRC/CCP4 file."""
-        mrcfile = mrc.MRC(filename)
+        mrcfile = mrc.MRC(filename, **kwargs)
```
orbeckst: Explicitly pull out kwargs
```suggestion
        mrcfile = mrc.MRC(filename, is_volume=kwargs.get("is_volume", None))
```

gridData/core.py
orbeckst: Or use the docs that you already wrote :-) (+ the versionadded)

gridData/mrc.py
```diff
@@ -43,6 +43,12 @@ class MRC(object):
     ----------
     filename : str (optional)
        input file (or stream), can be compressed
+    is_volume : bool (optional)
+       Force input file to be interpreted as a volume. If None (default),
+       use the MRC file header. If True, force the file to be read as a
+       volume. If False, the file is intepreted as a 2D image stack and
+       raise a ValueError.
+
```
orbeckst, line 51:
```suggestion
       
       .. versionadded:: 1.1.0

```

Check the rendered docs!

gridData/mrc.py
```diff
@@ -86,17 +92,22 @@ class MRC(object):
 
     """
 
-    def __init__(self, filename=None):
+    def __init__(self, filename=None, is_volume=None):
         self.filename = filename
         if filename is not None:
-            self.read(filename)
+            self.read(filename, is_volume)
```
orbeckst: Use kwargs explicitly, not implicitly.
```suggestion
            self.read(filename, is_volume=is_volume)
```

gridData/tests/test_mrc.py
```diff
@@ -113,6 +113,17 @@ def test_triclinic_ValueError():
                        "supported, not"):
         Grid(datafiles.MRC_EMD3001, file_format="MRC")
 
+def test_mrcfile_volume_check():
+    with pytest.raises(ValueError, match="is not a volumetric density"):
+        Grid(datafiles.ISPG_0)
+
+def test_mrcfile_volume_force():
+    Grid(datafiles.ISPG_0, is_volume=True)
```
orbeckst: `assert` something about the data

orbeckst (changes requested):
Thank you for making the changes. Please see comments.

Please also 
- [x] add an entry to CHANGELOG under Enhancements for 1.1.0
- [x] add yourself to AUTHORS
- [ ] check that the docs render correctly (follow the link under the [docs deployment check](https://github.com/MDAnalysis/GridDataFormats/actions/runs/21006546546/job/60390262269?pr=149))

EDIT: see also discussion below to use `assume_volumetric` instead of `is_volume`.

gridData/mrc.py
```diff
@@ -86,17 +92,22 @@ class MRC(object):
...
         if filename is not None:
             self.filename = filename
         with mrcfile.open(filename) as mrc:
-            if not mrc.is_volume():                           #pragma: no cover
+            if is_volume is None:
+                is_volume = mrc.is_volume()
+            elif is_volume:
+                # non 3D volumes should always fail, regardless of is_volume value
+                is_volume = mrc.data is not None and len(mrc.data.shape) == 3
+            if not is_volume:                           #pragma: no cover
```
PlethoraChutney: A question I had, never having used coverage.py before. Do we still need this `pragma: no cover`?

orbeckst: No, because you added an explicit test (which is great!).

Remove it. Good catch!

PlethoraChutney:
I'll try to put this all together later today. I was thinking though, it feels odd to have `Grid(filename, is_volume=False)` be a more verbose `raise ValueError`. How would we feel about changing to `coerce_volumetric: bool` default `False`?

BradyAJohnston:
Maybe not `coerce` but `assume_volumetric` ? We aren't changing the data at all just changing our assumptions about it.

orbeckst:
We can use `assume_volumetric=False` as default (as before) and `True` to do the same as `is_volume=True`. (`None` will just do the same as `False`, which is fine)

PlethoraChutney:
OK, latest commit changed to an explicit `assume_volumetric` parameter, and docs look alright to me.

CHANGELOG
```diff
@@ -28,6 +28,8 @@ The rules for this file:
   * `Grid` now accepts binary operations with any operand that can be 
     broadcasted to the grid's shape according to `numpy` broadcasting rules
     (PR #142)
+  * `Grid` now allows forcing MRC/CCP4 maps to be read as volumes even when
+    the header indicates they are stacks of 2D images.
```
orbeckst:
```suggestion
    the header indicates they are stacks of 2D images. (#150, PR #149)
```

gridData/mrc.py
```diff
@@ -86,17 +92,23 @@ class MRC(object):
...
 
-    def read(self, filename):
+    def read(self, filename, assume_volumetric=False):
         """Populate the instance from the MRC/CCP4 file *filename*."""
         if filename is not None:
             self.filename = filename
         with mrcfile.open(filename) as mrc:
-            if not mrc.is_volume():                           #pragma: no cover
+            if assume_volumetric:
+                # non 3D volumes should always fail, regardless of is_volume value
```
orbeckst:
```suggestion
                # non 3D volumes should always fail, regardless of assume_volumetric value
```

gridData/core.py
```diff
@@ -545,32 +554,32 @@ def load(self, filename, file_format=None):
             # are not really a file
             raise IOError(errno.ENOENT, "file not found", filename)
         loader = self._get_loader(filename, file_format=file_format)
-        loader(filename)
+        loader(filename, **kwargs)
 
-    def _load_python(self, filename):
+    def _load_python(self, filename, assume_volumetric=False):
```
orbeckst: What I don't quite like is that we are listing parameters here that are not relevant. Let's swallow parameters if we don't care and instead control the parameters at `load()`.
```suggestion
    def _load_python(self, filename, **kwargs):
```

gridData/core.py
```diff
@@ -532,7 +541,7 @@ def _load(
                 "grid={0} edges={1} origin={2} delta={3}".format(
                     grid, edges, origin, delta))
 
-    def load(self, filename, file_format=None):
+    def load(self, filename, file_format=None, **kwargs):
```
orbeckst:
```suggestion
    # NOTE: keep loader kwargs in sync between load() and __init__()
    def load(self, filename, file_format=None, assume_volumetric=False):
```

Be explicit here to catch user errors.

gridData/core.py
```diff
@@ -545,32 +554,32 @@ def load(self, filename, file_format=None):
             # are not really a file
             raise IOError(errno.ENOENT, "file not found", filename)
         loader = self._get_loader(filename, file_format=file_format)
-        loader(filename)
+        loader(filename, **kwargs)
```
orbeckst:
```suggestion
        loader(filename, assume_volumetric=assume_volumetric)
```

gridData/core.py
```diff
@@ -545,32 +554,32 @@ def load(self, filename, file_format=None):
...
-    def _load_python(self, filename):
+    def _load_python(self, filename, assume_volumetric=False):
         with open(filename, 'rb') as f:
             saved = pickle.load(f)
         self._load(grid=saved['grid'],
                    edges=saved['edges'],
                    metadata=saved['metadata'])
 
-    def _load_mrc(self, filename):
+    def _load_mrc(self, filename, assume_volumetric=False):
```
orbeckst: Swallow any kwargs that we do not care about (non, so far)
```suggestion
    def _load_mrc(self, filename, assume_volumetric=False, **kwargs):
```

gridData/core.py
```diff
@@ -545,32 +554,32 @@ def load(self, filename, file_format=None):
...
-        mrcfile = mrc.MRC(filename)
+        mrcfile = mrc.MRC(filename, assume_volumetric=assume_volumetric)
         grid, edges = mrcfile.histogramdd()
         self._load(grid=grid, edges=edges, metadata=self.metadata)
         # Store header for access from Grid object (undocumented)
         # https://github.com/MDAnalysis/GridDataFormats/pull/100#discussion_r782604833
         self._mrc_header = mrcfile.header.copy()
 
-    def _load_dx(self, filename):
+    def _load_dx(self, filename, assume_volumetric=False):
```
orbeckst:
```suggestion
    def _load_dx(self, filename, **kwargs):
```

gridData/core.py
```diff
@@ -545,32 +554,32 @@ def load(self, filename, file_format=None):
...
         """Initializes Grid from a OpenDX file."""
         dx = OpenDX.field(0)
         dx.read(filename)
         grid, edges = dx.histogramdd()
         self._load(grid=grid, edges=edges, metadata=self.metadata)
 
-    def _load_plt(self, filename):
+    def _load_plt(self, filename, assume_volumetric=False):
```
orbeckst:
```suggestion
    def _load_plt(self, filename, **kwargs):
```

orbeckst (approved):
Looks very good — I just want to make changes to how we pass arguments from `load()` to `_load_xxx()` private methods.

gridData/core.py
```diff
@@ -93,6 +93,15 @@ class Grid(object):
       format fails. The default is ``None`` and normally the file
       format is guessed from the file extension.
 
+    assume_volumetric : bool (optional)
+      If ``False`` (default), check the file header to determine whether
+      the data in `grid` is a 3D volume. If ``True``, assume `grid` is volumetric.
+
+      .. Note:: `assume_volumetric` only has an effect when loading
+         MRC/CCP4 files. See :class:`gridData.MRC.MRC`
```
orbeckst:
```suggestion
         MRC/CCP4 files. See :class:`gridData.mrc.MRC`
```

orbeckst:
Thank you for your contribution @PlethoraChutney — nicely done.

Thank you for your great comments @BradyAJohnston !

PR #159, opened 2026-01-22
ensure testing of built version after deployment

### LINKED ISSUE #158: testing in deployment workflow does not check for the correct version
The current deployment workflow tries to test the uploaded pypi (or testpypi) package by downloading the latest package, installing it, and running the test suite.

However, downloading "the latest package" can fail in multiple ways
1. The previous package is downloaded because pypi was not updated quickly enough. This may either lead to failures or false positives.
2. When testing with pre-release versions such as 1.1.0rc1, `pip install GridDataFormats` never considers pre-releases and always installs the latest stable so that we cannot properly test.

We can add a heuristic delay but that also addresses (1). A better approach would be to communicate the version number between workflow steps and then have pip install the exact version. Add either delay or retry functionality.

orbeckst:
- fix #158
- remember version that was built
- ensure that we download the built version from TestPyPi or PyPi
- built with AI assistance (cursor) (EDIT: removed the majority of the generated code, remaining code is majority human...)


orbeckst:
Deployment workflow for testpypi worked correctly and pulled the 1.1.1rc0 version. Note that it was tagged as 1.1.1c but the whole workflow succeeded, despite the build-renaming of the version.

https://github.com/MDAnalysis/GridDataFormats/actions/runs/21269343790/job/61215929498

orbeckst:
For PyPi I got an installation failure for the macos-latest: https://github.com/MDAnalysis/GridDataFormats/actions/runs/21269454572/job/61216293028

*Install from PyPi*
```
  python -m pip install --upgrade pip
  pip install "GridDataFormats[test]==1.1.1rc0"
  shell: /bin/bash -l {0}
  env:
    pythonLocation: /Users/runner/hostedtoolcache/Python/3.14.2/arm64
    PKG_CONFIG_PATH: /Users/runner/hostedtoolcache/Python/3.14.2/arm64/lib/pkgconfig
    Python_ROOT_DIR: /Users/runner/hostedtoolcache/Python/3.14.2/arm64
    Python2_ROOT_DIR: /Users/runner/hostedtoolcache/Python/3.14.2/arm64
    Python3_ROOT_DIR: /Users/runner/hostedtoolcache/Python/3.14.2/arm64
Requirement already satisfied: pip in /Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/site-packages (25.3)
ERROR: Ignored the following yanked versions: 1.0.0
ERROR: Could not find a version that satisfies the requirement GridDataFormats==1.1.1rc0 (from versions: 0.2.2, 0.2.3, 0.2.4, 0.2.5, 0.3.0, 0.3.1, 0.3.2, 0.3.3, 0.4.0, 0.4.1, 0.5.0, 0.6.0, 0.7.0, 1.0.1, 1.0.2, 1.1.0rc2, 1.1.0)
ERROR: No matching distribution found for GridDataFormats==1.1.1rc0
Error: Process completed with exit code 1.
```

The previous step succeeded and said that the version was on PyPi:

*Wait for version to be available on PyPi*
```
Available versions: ['0.4.1', '0.5.0', '0.6.0', '0.7.0', '1.0.0', '1.0.1', '1.0.2', '1.1.0', '1.1.0rc2', '1.1.1rc0']
✓ Version 1.1.1rc0 is available on PyPI
Version 1.1.1rc0 is now available on PyPI
```


Manual testing: the version exists https://pypi.org/project/GridDataFormats/1.1.1rc0/ and I can manually install (on macOS) with
```
$ pip install GridDataFormats[test]==1.1.1rc0

Successfully installed GridDataFormats-1.1.1rc0
```

Perhaps this indicates that there are still some race conditions or caching issues. 

Not sure how to debug/test. Perhaps just check if it works with another release, to check that it's not a PyPi thing (vs TestPyPi).

.github/actions/wait-for-pypi-version/action.yaml
```diff
@@ -0,0 +1,86 @@
...
+    - name: Wait for version to be available
+      shell: bash
+      env:
+        REPOSITORY: ${{ inputs.repository }}
+        PACKAGE: ${{ inputs.package }}
+        VERSION: ${{ inputs.version }}
+        MAX_ATTEMPTS: ${{ inputs.max_attempts }}
+        WAIT_SECONDS: ${{ inputs.wait_seconds }}
+      run: |
+        if [ "$REPOSITORY" = "testpypi" ]; then
```
IAlibay: Just a brief initial look - it feels like most of this should be doable directly as a python call.
Switchin between bash and python seems less efficient than having more os calls.

orbeckst: Yes, good point!

it’s probably also nicer to just write a Python script instead of inline code.

orbeckst:
Comment from @IAlibay on Discord:

> I'll be honest, for the sake of stability I just add a step where I ask the whole thing to sleep for a few minutes.
> We're never in such a hurry during a release that sleep(300) doesn't end up having a lower mainteannce / stability burden than trying to poll during that time.

Fair! Perhaps just replace the whole "wait until ready" action with `sleep(x)` but keep the explicit version code.

.github/workflows/deploy.yaml
```diff
@@ -143,10 +163,20 @@ jobs:
...
+      - name: Checkout repository for actions
+        uses: actions/checkout@v4
+
+      - name: Wait for version to be available on TestPyPI
+        uses: ./.github/actions/wait-for-pypi-version
+        with:
+          repository: testpypi
+          package: GridDataFormats
+          version: ${{ needs.build.outputs.version }}
+
```
orbeckst: For simpler version, replace custom action with a `sleep`, e.g. `sleep 300`.

.github/actions/wait-for-pypi-version/action.yaml
orbeckst: Do you prefer "inline python" over "Python script that is stored somewhere in the repo"? If script then where is a good place to put it? Under .github/bin or .github/utils .... open for suggestions @IAlibay

orbeckst: (I'll replace it with sleep 60 or something like that for right now but would be generally interested in best practices advice.)

.github/workflows/deploy.yaml
orbeckst: now uses `sleep 30`

orbeckst: The fancy version actually works pretty well, see https://github.com/Becksteinlab/multibind/blob/master/.github/actions/wait-for-pypi-version/action.yaml and https://github.com/Becksteinlab/multibind/blob/7c93f605d99ff67d9adef890c9302bccd2caa1b5/.github/workflows/deploy.yaml#L171-L176

orbeckst:
Automatic deployment and testing is working with the 30s delay, tested with 1.1.1rc1
- [x] TestPypi https://github.com/MDAnalysis/GridDataFormats/actions/runs/24222925255 and https://test.pypi.org/project/GridDataFormats/1.1.1rc1/
- [x] PyPi  https://github.com/MDAnalysis/GridDataFormats/actions/runs/24223020174/job/70718457380 and https://pypi.org/project/GridDataFormats/1.1.1rc1/

@IAlibay are you happy with the current changes?

.github/actions/wait-for-pypi-version/action.yaml
orbeckst: replaced with `sleep 30`

IAlibay (approved):
This looks good to me!

PR #163, opened 2026-03-06
Api openvdb update

spyke7:
This PR is related to #148 PR.
In this I have tried to fix #161  and #160 , only regarding `openvdb`.

Here are the changes- 
- added convert_to function that returns openvdb.GridBase
- inisde `__init__` we check if given grid is native or not, and depending on it we define self.vdb_grid
- In `OpenVDB.py`, `_extract_from_vdb_grid` is made to extract origin, delta and grid from given native object
- It contains some upgrade and remaining tests for OpenVDB.py (#148)

@orbeckst pls check!

PR #148, opened 2025-12-27
added a simple program to export files in .vdb format

spyke7:
Hi @orbeckst 
I have added `OpenVDB.py` inside gridData that simply export files in `.vdb` format. Also I have added `test_vdb.py` inside tests and it successfully passes.
fix #141 

Required Libraries - 
openvdb
- `conda install -c conda-forge openvdb`

There are many things that need to be updated like docs, etc, but I have just provided the file and test so that you can review it, and I can fix the problems. Please let me know if anything needs to be changed and updated.

PR #164, opened 2026-03-11
API update for OpenDX and MRC

### LINKED ISSUE #161: API-interoperability: Grid.convert_to()
Implement the conversion from a Grid to a native object as described in the parent issue #160 for

- mrcfile: `mrcfile.MrcFile`
- OpenDX:  `gridData.OpenDX.field` (i.e., `self`)
- openvdb (once we have #141 ): `OpenVDB.OpenVDBField`

spyke7:
This is regarding #161 

Implemented from_grid() a static method and native (property) inside OpenDX.py and mrc.py

- [x] Just simply added those two functions inside files
- [x] register converters
- [x] change _export_<formats>
- [x] write tests

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -19,9 +19,11 @@ The rules for this file:
 
   Enhancements 
 
-  * `from_grid()` and `native` functions are added inside mrc and OpenDX,
-    which is used by `Grid._export_<FORMAT>` and a new `convert_to()` is added
-    (issue #161, PR #164)
+  * new API-level interoperability features (#160)
+    * new `Grid.convert_to()` method to create the native object (#161, PR #164)
+    * `mrc.MRC`, `OpenDX.field`: new `from_grid()` class method to create the
+      correpsonding GridDataFormats wrapper from a `Grid`; the new attribute `native`
+      contains the native object (#161, PR #164)
 
   Fixes
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -22,7 +22,7 @@ The rules for this file:
   * new API-level interoperability features (#160)
     * new `Grid.convert_to()` method to create the native object (#161, PR #164)
     * `mrc.MRC`, `OpenDX.field`: new `from_grid()` class method to create the
-      correpsonding GridDataFormats wrapper from a `Grid`; the new attribute `native`
+      corresponding GridDataFormats wrapper from a `Grid`; the new attribute `native`
       contains the native object (#161, PR #164)
 
   Fixes
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,9 +15,9 @@ The rules for this file:
 -------------------------------------------------------------------------------
 ??/??/???? orbeckst, spyke7
 
- * 1.1.1
+ * 1.2.0
 
-  Changes
+  Enhancements 
 
   * `from_grid()` and `native` functions are added inside mrc and OpenDX,
     which is used by `Grid._export_<FORMAT>` and a new `convert_to()` is added
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,11 +13,17 @@ The rules for this file:
   * accompany each entry with github issue/PR number (Issue #xyz)
 
 -------------------------------------------------------------------------------
-??/??/???? orbeckst
+??/??/???? orbeckst, spyke7
 
  * 1.1.1
 
- Fixes
+  Changes
+
+  * `from_grid()` and `native` functions are added inside mrc and OpenDX,
+    which is used by `Grid._export_<FORMAT>` and a new `convert_to()` is added
+    (issue #161, PR #164)
+
+  Fixes
 
 
 01/22/2026 IAlibay, ollyfutur, conradolandia, orbeckst, PlethoraChutney,
--- a/gridData/OpenDX.py
+++ b/gridData/OpenDX.py
@@ -211,31 +211,42 @@ import warnings
 # Python 2/3 compatibility (see issue #99)
 # and https://bugs.python.org/issue30012
 import sys
-if sys.version_info >= (3, ):
+
+if sys.version_info >= (3,):
+
     def _gzip_open(filename, mode="rt"):
         return gzip.open(filename, mode)
+
 else:
+
     def _gzip_open(filename, mode="rt"):
         return gzip.open(filename)
+
+
 del sys
+
 
 class DXclass(object):
     """'class' object as defined by OpenDX"""
-    def __init__(self,classid):
+
+    def __init__(self, classid):
         """id is the object number"""
         self.id = classid  # serial number of the object
-        self.name = None   # name of the DXclass
-        self.component = None   # component type
-        self.D = None      # dimensions
+        self.name = None  # name of the DXclass
+        self.component = None  # component type
+        self.D = None  # dimensions
 
     def write(self, stream, optstring="", quote=False):
         """write the 'object' line; additional args are packed in string"""
         classid = str(self.id)
-        if quote: classid = '"'+classid+'"'
+        if quote:
+            classid = '"' + classid + '"'
         # Only use a *single* space between tokens; both chimera's and pymol's DX parser
         # does not properly implement the OpenDX specs and produces garbage with multiple
         # spaces. (Chimera 1.4.1, PyMOL 1.3)
-        to_write = 'object '+classid+' class '+str(self.name)+' '+optstring+'\n'
+        to_write = (
+            "object " + classid + " class " + str(self.name) + " " + optstring + "\n"
+        )
         self._write_line(stream, to_write)
 
     @staticmethod
@@ -246,15 +257,15 @@ class DXclass(object):
         stream.write(line)
 
     def read(self, stream):
-        raise NotImplementedError('Reading is currently not supported.')
+        raise NotImplementedError("Reading is currently not supported.")
 
-    def ndformat(self,s):
+    def ndformat(self, s):
         """Returns a string with as many repetitions of s as self
         has dimensions (derived from shape)"""
         return s * len(self.shape)
 
     def __repr__(self):
-        return '<OpenDX.'+str(self.name)+' object, id='+str(self.id)+'>'
+        return "<OpenDX." + str(self.name) + " object, id=" + str(self.id) + ">"
 
 
 class gridpositions(DXclass):
@@ -264,17 +275,18 @@ class gridpositions(DXclass):
     origin    coordinates of the centre of the grid cell with index 0,0,...,0
     delta     DxD array describing the deltas
     """
-    def __init__(self,classid,shape=None,origin=None,delta=None,**kwargs):
-        if shape is None or origin is None or delta is None:
-            raise ValueError('all keyword arguments are required')
-        self.id = classid
-        self.name = 'gridpositions'
-        self.component = 'positions'
-        self.shape = numpy.asarray(shape)      # D dimensional shape
-        self.origin = numpy.asarray(origin)    # D vector
-        self.rank = len(self.shape)            # D === rank
 
-        self.delta = numpy.asarray(delta)      # DxD array of grid spacings
+    def __init__(self, classid, shape=None, origin=None, delta=None, **kwargs):
+        if shape is None or origin is None or delta is None:
+            raise ValueError("all keyword arguments are required")
+        self.id = classid
+        self.name = "gridpositions"
+        self.component = "positions"
+        self.shape = numpy.asarray(shape)  # D dimensional shape
+        self.origin = numpy.asarray(origin)  # D vector
+        self.rank = len(self.shape)  # D === rank
+
+        self.delta = numpy.asarray(delta)  # DxD array of grid spacings
         # gridDataFormats  actually provides a simple 1D array with the deltas because only
         # regular grids are used but the following is a reminder that OpenDX should be able
         # to handle more complicated volume elements
@@ -283,40 +295,49 @@ class gridpositions(DXclass):
         if self.delta.shape != (self.rank, self.rank):
             # check OpenDX specs for irreg spacing if we want to implement
             # anything more complicated
-            raise NotImplementedError('Only regularly spaced grids allowed, '
-                                      'not delta={}'.format(self.delta))
+            raise NotImplementedError(
+                "Only regularly spaced grids allowed, "
+                "not delta={}".format(self.delta)
+            )
+
     def write(self, stream):
         super(gridpositions, self).write(
-            stream, ('counts '+self.ndformat(' %d')) % tuple(self.shape))
-        self._write_line(stream, 'origin %f %f %f\n' % tuple(self.origin))
+            stream, ("counts " + self.ndformat(" %d")) % tuple(self.shape)
+        )
+        self._write_line(stream, "origin %f %f %f\n" % tuple(self.origin))
         for delta in self.delta:
             self._write_line(
-                stream, ('delta ' +
-                         self.ndformat(' {:.7g}').format(*delta) +
-                         '\n'))
+                stream, ("delta " + self.ndformat(" {:.7g}").format(*delta) + "\n")
+            )
 
     def edges(self):
         """Edges of the grid cells, origin at centre of 0,0,..,0 grid cell.
 
         Only works for regular, orthonormal grids.
         """
-        return [self.delta[d,d] * numpy.arange(self.shape[d]+1) + self.origin[d]\
-                - 0.5*self.delta[d,d]     for d in range(self.rank)]
+        return [
+            self.delta[d, d] * numpy.arange(self.shape[d] + 1)
+            + self.origin[d]
+            - 0.5 * self.delta[d, d]
+            for d in range(self.rank)
+        ]
 
 
 class gridconnections(DXclass):
     """OpenDX gridconnections class"""
-    def __init__(self,classid,shape=None,**kwargs):
+
+    def __init__(self, classid, shape=None, **kwargs):
         if shape is None:
-            raise ValueError('all keyword arguments are required')
+            raise ValueError("all keyword arguments are required")
         self.id = classid
-        self.name = 'gridconnections'
-        self.component = 'connections'
-        self.shape = numpy.asarray(shape)      # D dimensional shape
+        self.name = "gridconnections"
+        self.component = "connections"
+        self.shape = numpy.asarray(shape)  # D dimensional shape
 
     def write(self, stream):
         super(gridconnections, self).write(
-            stream, ('counts '+self.ndformat(' %d')) % tuple(self.shape))
+            stream, ("counts " + self.ndformat(" %d")) % tuple(self.shape)
+        )
 
 
 class array(DXclass):
@@ -327,22 +348,23 @@ class array(DXclass):
     .. _Array Objects:
        https://web.archive.org/web/20080808140524/http://opendx.sdsc.edu/docs/html/pages/usrgu068.htm#Header_440
     """
+
     #: conversion from :attr:`numpy.dtype.name` to closest OpenDX array type
     #: (round-tripping is not guaranteed to produce identical types); not all
     #: types are supported (e.g., strings are missing)
     np_types = {
-        "uint8": "byte",         # DX "unsigned byte" equivalent
+        "uint8": "byte",  # DX "unsigned byte" equivalent
         "int8": "signed byte",
         "uint16": "unsigned short",
-        "int16": "short",         # DX "signed short" equivalent
+        "int16": "short",  # DX "signed short" equivalent
         "uint32": "unsigned int",
-        "int32": "int",           # DX "signed int"   equivalent
-        "uint64": "unsigned int", # not explicit in DX, for compatibility
-        "int64": "int",           # not explicit in DX, for compatibility
+        "int32": "int",  # DX "signed int"   equivalent
+        "uint64": "unsigned int",  # not explicit in DX, for compatibility
+        "int64": "int",  # not explicit in DX, for compatibility
         # "hyper",                # ?
-        "float32": "float",       # default
+        "float32": "float",  # default
         "float64": "double",
-        "float16": "float",       # float16 not available in DX, use float
+        "float16": "float",  # float16 not available in DX, use float
         # numpy "float128 not available, raise error
         # "string" not automatically supported
     }
@@ -360,13 +382,12 @@ class array(DXclass):
         "int": "int32",
         "signed int": "int32",
         # "hyper",                # ?
-        "float": "float32",       # default
+        "float": "float32",  # default
         "double": "float64",
         # "string" not automatically supported
     }
 
-    def __init__(self, classid, array=None, type=None, typequote='"',
-                 **kwargs):
+    def __init__(self, classid, array=None, type=None, typequote='"', **kwargs):
         """
         Parameters
         ----------
@@ -388,29 +409,35 @@ class array(DXclass):
              DX type
         """
         if array is None:
-            raise ValueError('array keyword argument is required')
+            raise ValueError("array keyword argument is required")
         self.id = classid
-        self.name = 'array'
-        self.component = 'data'
+        self.name = "array"
+        self.component = "data"
         # detect type https://github.com/MDAnalysis/GridDataFormats/issues/35
         if type is None:
             self.array = numpy.asarray(array)
             try:
                 self.type = self.np_types[self.array.dtype.name]
             except KeyError:
-                warnings.warn(("array dtype.name = {0} can not be automatically "
-                               "converted to a DX array type. Use the 'type' keyword "
-                               "to manually specify the correct type.").format(
-                                   self.array.dtype.name))
+                warnings.warn(
+                    (
+                        "array dtype.name = {0} can not be automatically "
+                        "converted to a DX array type. Use the 'type' keyword "
+                        "to manually specify the correct type."
+                    ).format(self.array.dtype.name)
+                )
                 self.type = self.array.dtype.name  # will raise ValueError on writing
         else:
             try:
                 self.array = numpy.asarray(array, dtype=self.dx_types[type])
             except KeyError:
-                raise ValueError(("DX type {0} cannot be converted to an "
-                                  "appropriate numpy dtype. Available "
-                                  "types are: {1}".format(type,
-                                                          list(self.dx_types.values()))))
+                raise ValueError(
+                    (
+                        "DX type {0} cannot be converted to an "
+                        "appropriate numpy dtype. Available "
+                        "types are: {1}".format(type, list(self.dx_types.values()))
+                    )
+                )
             self.type = type
         self.typequote = typequote
 
@@ -429,32 +456,38 @@ class array(DXclass):
 
         """
         if self.type not in self.dx_types:
-            raise ValueError(("DX type {} is not supported in the DX format. \n"
-                              "Supported valus are: {}\n"
-                              "Use the type=<type> keyword argument.").format(
-                                  self.type, list(self.dx_types.keys())))
-        typelabel = (self.typequote+self.type+self.typequote)
-        super(array, self).write(stream, 'type {0} rank 0 items {1} data follows'.format(
-            typelabel, self.array.size))
+            raise ValueError(
+                (
+                    "DX type {} is not supported in the DX format. \n"
+                    "Supported valus are: {}\n"
+                    "Use the type=<type> keyword argument."
+                ).format(self.type, list(self.dx_types.keys()))
+            )
+        typelabel = self.typequote + self.type + self.typequote
+        super(array, self).write(
+            stream,
+            "type {0} rank 0 items {1} data follows".format(typelabel, self.array.size),
+        )
 
         # grid data, serialized as a C array (z fastest varying)
         # (flat iterator is equivalent to: for x: for y: for z: grid[x,y,z])
         # VMD's DX reader requires exactly 3 values per line
         fmt_string = "{:d}"
-        if (self.array.dtype.kind == 'f' or self.array.dtype.kind == 'c'):
+        if self.array.dtype.kind == "f" or self.array.dtype.kind == "c":
             precision = numpy.finfo(self.array.dtype).precision
-            fmt_string = "{:."+"{:d}".format(precision)+"f}"
+            fmt_string = "{:." + "{:d}".format(precision) + "f}"
         values_per_line = 3
         values = self.array.flat
         while 1:
             try:
                 for i in range(values_per_line):
                     self._write_line(stream, fmt_string.format(next(values)) + "\t")
-                self._write_line(stream, '\n')
+                self._write_line(stream, "\n")
             except StopIteration:
-                self._write_line(stream, '\n')
+                self._write_line(stream, "\n")
                 break
         self._write_line(stream, 'attribute "dep" string "positions"\n')
+
 
 class field(DXclass):
     """OpenDX container class
@@ -466,9 +499,10 @@ class field(DXclass):
     :meth:`add`.
 
     """
+
     # perhaps this should not derive from DXclass as those are
     # objects in field but a field cannot contain itself
-    def __init__(self,classid='0',components=None,comments=None):
+    def __init__(self, classid="0", components=None, comments=None):
         """OpenDX object, which is build from a list of components.
 
         Parameters
@@ -511,22 +545,24 @@ class field(DXclass):
 
         """
         if components is None:
-            components = dict(positions=None,connections=None,data=None)
+            components = dict(positions=None, connections=None, data=None)
         if comments is None:
-            comments = ['OpenDX written by gridData.OpenDX',
-                        'from https://github.com/MDAnalysis/GridDataFormats']
+            comments = [
+                "OpenDX written by gridData.OpenDX",
+                "from https://github.com/MDAnalysis/GridDataFormats",
+            ]
         elif type(comments) is not list:
             comments = [str(comments)]
-        self.id = classid       # can be an arbitrary string
-        self.name = 'field'
-        self.component = None   # cannot be a component of a field
+        self.id = classid  # can be an arbitrary string
+        self.name = "field"
+        self.component = None  # cannot be a component of a field
         self.components = components
-        self.comments= comments
+        self.comments = comments
 
     @staticmethod
     def from_grid(grid, type=None, typequote='"', **kwargs):
         """Create OpenDX field from Grid.
-    
+
         Parameters
         ----------
         grid : Grid
@@ -534,7 +570,7 @@ class field(DXclass):
         typequote : str, optional
         **kwargs
             Additional keyword arguments (currently unused)
-        
+
         Returns
         -------
         field
@@ -552,13 +588,13 @@ class field(DXclass):
         for k in grid.metadata:
             comments.append("   " + str(k) + " = " + str(grid.metadata[k]))
         comments.append("(Note: the VMD dx-reader chokes on comments below this line)")
-        
+
         components = dict(
             positions=gridpositions(1, grid.grid.shape, grid.origin, grid.delta),
             connections=gridconnections(2, grid.grid.shape),
             data=array(3, grid.grid, type=type, typequote=typequote),
         )
-        dx_field = field('density', components=components, comments=comments)
+        dx_field = field("density", components=components, comments=comments)
         return dx_field
 
     @property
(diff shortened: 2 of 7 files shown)


spyke7:
@orbeckst  pls check this kindly

orbeckst (changes requested):
This is already looking pretty good. Please also add doc strings and update CHANGELOG. Will have a closer look once this is done. Thank you!

gridData/tests/test_dx.py
```diff
@@ -90,3 +90,26 @@ def test_delta_precision(tmpdir):
...
+    data = np.ones((5, 5, 5), dtype=np.float32)
+    g = Grid(data, origin=[1.0, 2.0, 3.0], delta=[0.5, 0.5, 0.5])
+    g.metadata['name'] = 'test_density'
+    g.metadata['author'] = 'test_user'
+    
+    dx_field = gridData.OpenDX.field.from_grid(g)
+    
+    assert isinstance(dx_field, gridData.OpenDX.field)
+    
+    assert any('test_density' and 'test_user' in str(c) for c in dx_field.comments)
```
BradyAJohnston: This is a logic error. `'test_density' and 'test_user' in str(c)` evaluates as `'test_user' in str(c)` due to Python's `and` short-circuit semantics — `'test_density'` is truthy (not `None` / empty string), so it's ignored. Should be:
```py                                              
assert any('test_density' in str(c) for c in dx_field.comments)                                                                 
assert any('test_user' in str(c) for c in dx_field.comments)
```

gridData/core.py
```diff
@@ -212,6 +212,12 @@ class Grid(object):
 
     #: Default format for exporting with :meth:`export`.
     default_format = 'DX'
+    
+    converter = {
+        'MRC': mrc.MRC.from_grid, 
+        'VDB': None,
```
BradyAJohnston: VDB is going to be implemented but it should be removed for now. Currently the `self.converter[]()` has potential to try and call `None()` so best to just not include until the VDB support is added (or add a guard for such in the later usage).

gridData/core.py
```diff
@@ -598,6 +604,32 @@ def _load_plt(self, filename, **kwargs):
         grid, edges = g.histogramdd()
         self._load(grid=grid, edges=edges, metadata=self.metadata)
 
+    def convert_to(self, format_specifier, tolerance=None, **kwargs):
```
BradyAJohnston: The tolerance is currently not used at all. Should be removed and left for the PR that introduces VDB.

gridData/mrc.py
```diff
@@ -101,6 +101,40 @@ def __init__(self, filename=None, assume_volumetric=False):
         if filename is not None:
             self.read(filename, assume_volumetric=assume_volumetric)
 
+    @staticmethod
```
BradyAJohnston: Should these be classmethods instead of staticmethods?

BradyAJohnston (changes requested):

gridData/tests/test_dx.py
BradyAJohnston: Actually to get what you are doing there, you might need 

```py

assert any(('test_density' in str(c) and 'test_user' in str(c)) for c in dx_field.comments)
```

If you need them both to be true for a single comment.

gridData/mrc.py
spyke7: In the issue it was staticmethod and it's fine for basic usage as well. But classmethod is more good, if somebody subclasses mrc/dx
Should I change it?

BradyAJohnston: I would have thought it makes more sense to make it classmethod, but @orbeckst was specific about it being static method in the issue. @orbeckst I'm interested why?

CHANGELOG
```diff
@@ -13,11 +13,17 @@ The rules for this file:
   * accompany each entry with github issue/PR number (Issue #xyz)
 
 -------------------------------------------------------------------------------
-??/??/???? orbeckst
+??/??/???? orbeckst, spyke7
 
  * 1.1.1
 
- Fixes
+  Changes
```
orbeckst: *Enhancements* (not changes)

CHANGELOG
```diff
@@ -13,11 +13,17 @@ The rules for this file:
   * accompany each entry with github issue/PR number (Issue #xyz)
 
 -------------------------------------------------------------------------------
-??/??/???? orbeckst
+??/??/???? orbeckst, spyke7
 
  * 1.1.1
```
orbeckst: Change to 1.2.0 (under semantic versioning, we must increase minor version for new features)

gridData/OpenDX.py
```diff
@@ -511,24 +545,69 @@ def __init__(self,classid='0',components=None,comments=None):
...
+        grid : Grid
+        type : str, optional
+        typequote : str, optional
+        **kwargs
+            Additional keyword arguments (currently unused)
+
+        Returns
+        -------
+        field
+            OpenDX field wrapper
```
orbeckst, line 586: add a `.. versionadded:: 1.2.0` (with two blank lines above)

gridData/OpenDX.py
```diff
@@ -511,24 +545,69 @@ def __init__(self,classid='0',components=None,comments=None):
...
+
+    @staticmethod
+    def from_grid(grid, type=None, typequote='"', **kwargs):
+        """Create OpenDX field from Grid.
+
+        Parameters
+        ----------
+        grid : Grid
+        type : str, optional
+        typequote : str, optional
```
orbeckst, line 575: add minimal explanations (can copy from other docs)

gridData/OpenDX.py
```diff
@@ -511,24 +545,69 @@ def __init__(self,classid='0',components=None,comments=None):
...
+            positions=gridpositions(1, grid.grid.shape, grid.origin, grid.delta),
+            connections=gridconnections(2, grid.grid.shape),
+            data=array(3, grid.grid, type=type, typequote=typequote),
+        )
+        dx_field = field("density", components=components, comments=comments)
+        return dx_field
+
+    @property
+    def native(self):
+        """Return native object"""
```
orbeckst: - Add a sentence "The "native" object is the :class:`gridData.OpenDX.field` itself."
- add a `.. versionadded:: 1.2.0` (with two blank lines above)

gridData/core.py
orbeckst: You can comment out this line.

gridData/core.py
```diff
@@ -598,6 +604,32 @@ def _load_plt(self, filename, **kwargs):
...
+        Parameters
+        ----------
+        format_specifier : str 
+        
+        tolerance : float (default is None) - for OpenVDB
+            
+        Returns
+        -------
+        native object
+
```
orbeckst, line 646: add a `.. versionadded:: 1.2.0` (with two blank lines above)

gridData/core.py
```diff
@@ -598,6 +604,32 @@ def _load_plt(self, filename, **kwargs):
         grid, edges = g.histogramdd()
         self._load(grid=grid, edges=edges, metadata=self.metadata)
 
+    def convert_to(self, format_specifier, tolerance=None, **kwargs):
+        """Returns an instance of the native object for a given format
+        
+        Implemented formats:
+        
+        DX
+            :mod:`OpenDX`
```
orbeckst: link to the native object, i.e. `OpenDX.field`

gridData/mrc.py
```diff
@@ -101,6 +101,40 @@ def __init__(self, filename=None, assume_volumetric=False):
...
+
+        if hasattr(grid, "_mrc_header"):
+            mrc_obj.header = grid._mrc_header
+
+        return mrc_obj
+
+    @property
+    def native(self):
+        """Native object is the MRC wrapper itself."""
+        return self
```
orbeckst: no, this should be a `mrcfile.MrcFile` (see https://github.com/MDAnalysis/GridDataFormats/issues/161#issuecomment-4014683563 )

orbeckst: Also add the versionadded here and docs.

gridData/mrc.py
```diff
@@ -101,6 +101,40 @@ def __init__(self, filename=None, assume_volumetric=False):
...
+        grid : Grid
+            Grid object to convert
+        **kwargs
+            Additional keyword arguments (currently unused)
+
+        Returns
+        -------
+        MRC
+            MRC wrapper object
+
```
orbeckst, line 123: add versionadded 1.2.0.

gridData/mrc.py
```diff
@@ -101,6 +101,40 @@ def __init__(self, filename=None, assume_volumetric=False):
         if filename is not None:
             self.read(filename, assume_volumetric=assume_volumetric)
 
+    @staticmethod
+    def from_grid(grid, **kwargs):
+        """Create MRC object from a Grid.
+
```
orbeckst, line 111: add text explaining that if `grid` was originally created from a mrcfile (and thus has the `Grid._mrc_header` attribute with the MRC header, then the MRC header will be copied into the `MRC` instance.

gridData/mrc.py
orbeckst: I did't know any better ;-). After some reading I also recommend going with `@classmethod` — if we need to work with inheritance or use it from another classmethod then classmethod will work better.

gridData/OpenDX.py
```diff
@@ -511,24 +545,69 @@ def __init__(self,classid='0',components=None,comments=None):
...
-        self.name = 'field'
-        self.component = None   # cannot be a component of a field
+        self.id = classid  # can be an arbitrary string
+        self.name = "field"
+        self.component = None  # cannot be a component of a field
         self.components = components
-        self.comments= comments
+        self.comments = comments
+
+    @staticmethod
```
orbeckst: change to `@classmethod` (see other discussion); don't forget to add `cls` as first argument

gridData/core.py
```diff
@@ -598,6 +604,32 @@ def _load_plt(self, filename, **kwargs):
...
+        MRC
+            :mod:`mrc` MRC/CCP4 format
```
orbeckst: link to the native object, i.e. `mrcfile.MrcFile`

gridData/tests/test_dx.py
```diff
@@ -90,3 +90,26 @@ def test_delta_precision(tmpdir):
...
+    # assert any('test_user' in str(c) for c in dx_field.comments)
```
orbeckst: remove commented out code

gridData/tests/test_dx.py
```diff
@@ -90,3 +90,26 @@ def test_delta_precision(tmpdir):
...
+    
+    assert_equal(dx_field.components['data'].array, data)
+    assert_equal(dx_field.components['positions'].origin, g.origin)
```
orbeckst: These may be floats so use [assert_allclose](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose).

gridData/tests/test_dx.py
```diff
@@ -90,3 +90,26 @@ def test_delta_precision(tmpdir):
...
+    
+def test_dx_native():
+    data = np.ones((5, 5, 5))
+    g = Grid(data, origin=[0, 0, 0], delta=[1, 1, 1])
+    
+    dx_field = gridData.OpenDX.field.from_grid(g)
+    assert dx_field.native is dx_field
```
orbeckst: This checks that it's the same object. Add another assertion that checks the type explicitly
```
assert isinstance(dx_field.native, OpenDX.field)
```

gridData/tests/test_mrc.py
```diff
@@ -142,6 +142,27 @@ def test_origin(self, grid, ccp4data):
...
+        assert_allclose(mrc_obj.origin, g.origin)
+        assert mrc_obj.rank == 3
+        
+    def test_mrc_native_property(self):
+        data = np.ones((3, 3, 3))
+        g = Grid(data, origin=[0, 0, 0], delta=[1, 1, 1])
+        
+        mrc_obj = mrc.MRC.from_grid(g)
+        
+        assert mrc_obj.native is mrc_obj
```
orbeckst: Needs to change as the native object should be the mrcfile.MrcFile.
- check type
- check some values

gridData/tests/test_mrc.py
```diff
@@ -142,6 +142,27 @@ def test_origin(self, grid, ccp4data):
...
+        
+    def test_mrc_from_grid(self):
+        data = np.arange(27, dtype=np.float32).reshape((3, 3, 3))
+        g = Grid(data, origin=[1.0, 2.0, 3.0], delta=[0.5, 0.5, 0.5])
+        
+        mrc_obj = mrc.MRC.from_grid(g)
+        
+        assert isinstance(mrc_obj, mrc.MRC)
+        
+        assert_equal(mrc_obj.array, data)
```
orbeckst: Given that the data are integers, assert_equal is technically correct but for consistency and robustness, also use assert_allclose, please.

orbeckst (changes requested):
Overall looking good but there are some necessary changes. In particular, the MRC.native should return a MrcFile.

Could you please also add a section to the [gridData.core docs](https://www.mdanalysis.org/GridDataFormats/gridData/core.html) about using `from_grid()` and `.native`? Have a look at the issue text for #161 and you can probably copy and paste from there. We just want to document somewhere for users that this API exists.

gridData/OpenDX.py
spyke7: yep

gridData/mrc.py
spyke7: For this do a new file need to be created?
or do a in-memory stream? as this mrcfile.MrcFile should be a file right? (self is `mrc_obj` that is returned in from_grid())

So native could be like this - 
```
stream = io.BytesIO()
self.write(stream)

return mrcfile.open(stream)
```

is this right? or am I missing something?

gridData/tests/test_dx.py
spyke7: > 
> ```python
> assert any(('test_density' in str(c) and 'test_user' in str(c)) for c in dx_field.comments)
> ```

This is giving error - 
```
>       assert any(('test_density' in str(c) and 'test_user' in str(c)) for c in dx_field.comments)
E       assert False
E        +  where False = any(<generator object test_dx_from_grid.<locals>.<genexpr> at 0x0000021CD04CF220>)

gridData\tests\test_dx.py:104: AssertionError
```


Either I can do them separately or I can do this - 
`assert any(('test_density' and 'test_user') in str(c) for c in dx_field.comments) `

spyke7:
@orbeckst @BradyAJohnston what to do exactly for `native` for mrc.py, as I am not able to understand how to return a mrcfie.MrcFile?
Please review other changes, but kept that part same as previous.

orbeckst:
Good point, MrcFile is tied to a file on disk. We could simply link it to a tmp file or maybe it works with StringIO? Or pass a filename in `convert_to()` and as optional kwarg? I have to think about this a little bit more and suggestions are welcome!

spyke7:
> Good point, MrcFile is tied to a file on disk. We could simply link it to a tmp file or maybe it works with StringIO?

inside write() the filename can be a BytesIO object, as 
```
if filename is not None:
            self.filename = filename
```
So ig it will work

Will also try the tempfile as well.

spyke7:
I used the tempfile at last cause of an error while using BytesIO
```
 def _open_file(self, name):
        """Open a file object to use as the I/O stream."""
>       self._iostream = open(name, self._mode + 'b')
```

`mrcfile` uses the open() function and it requires string not bytesIO

orbeckst:
@spyke7 thanks for doing the experiments. Could we use [mrcfile.mrcinterpreter.MrcInterpreter](https://mrcfile.readthedocs.io/en/stable/source/mrcfile.html#mrcfile.mrcinterpreter.MrcInterpreter) (instead of `MrcFile`) as it is built to take a stream?

CHANGELOG
```diff
@@ -13,11 +13,17 @@ The rules for this file:
...
 
- * 1.1.1
+ * 1.2.0
 
- Fixes
+  Enhancements 
+
+  * `from_grid()` and `native` functions are added inside mrc and OpenDX,
+    which is used by `Grid._export_<FORMAT>` and a new `convert_to()` is added
+    (issue #161, PR #164)
```
orbeckst: We can be a bit bolder here, it's a cool new feature!
```suggestion
  * new API-level interoperability features (#160)
    * new `Grid.convert_to()` method to create the native object (#161, PR #164)
    * `mrc.MRC`, `OpenDX.field`: new `from_grid()` class method to create the
      correpsonding GridDataFormats wrapper from a `Grid`; the new attribute `native`
      contains the native object (#161, PR #164)
```

gridData/mrc.py
```diff
@@ -101,6 +102,61 @@ def __init__(self, filename=None, assume_volumetric=False):
...
+        .. versionadded:: 1.2.0
+
+        """
+        fd, temp_path = tempfile.mkstemp(suffix=".mrc")
+
+        os.close(fd)
+        self.write(temp_path)
+
+        return mrcfile.open(temp_path)
```
orbeckst: This is a reasonable workaround but it's a bit clunky, we have to parse the data twice (once in write, then after open)  and we leave a tempfile around that will fill up the disk.

A slightly better solution would be to `self.write()` to a buffer and then have [`mrcfile.mrcinterpreter.MrcInterpreter`](https://mrcfile.readthedocs.io/en/stable/source/mrcfile.html#module-mrcfile.mrcinterpreter) directly read the buffer. That at least avoids the file on disk issue.

If that works then perhaps we can move almost all code from `write()` into native and the `write()` would just write the internal stream buffer to a file and in this way we only parse the data once with mrcfile.

orbeckst: Let's try at least the "slightly better" solution — if that works then we can use it for now.

gridData/tests/test_mrc.py
```diff
@@ -143,167 +179,191 @@ def test_origin(self, grid, ccp4data):
...
+
+    def test_mrc_native_property(self):
+        data = np.ones((3, 3, 3))
+        g = Grid(data, origin=[0, 0, 0], delta=[1, 1, 1])
+
+        mrc_wrapper = mrc.MRC.from_grid(g)
+        native_mrc = mrc_wrapper.native
+
+        assert isinstance(native_mrc, mrcfile.mrcfile.MrcFile)
+        np.testing.assert_allclose(native_mrc.data, data.T)
```
orbeckst, line 204: Why do we need to transpose data to make it agree?

gridData/tests/test_mrc.py
```diff
@@ -143,167 +179,191 @@ def test_origin(self, grid, ccp4data):
...
+
+        assert isinstance(mrc_obj, mrc.MRC)
+
+        assert_allclose(mrc_obj.array, data)
+        assert_allclose(mrc_obj.delta, np.diag(g.delta))
+        assert_allclose(mrc_obj.origin, g.origin)
+        assert mrc_obj.rank == 3
+
+    def test_mrc_native_property(self):
+        data = np.ones((3, 3, 3))
```
orbeckst: use a non-symmetric grid eg `arange(60).reshape(3,4,5)` ; with only ones you can miss issues with axes orientations.

gridData/tests/test_mrc.py
```diff
@@ -143,167 +179,191 @@ def test_origin(self, grid, ccp4data):
...
+        g = Grid(data, origin=[0, 0, 0], delta=[1, 1, 1])
```
orbeckst: Vary`delta = [1., 3., 0.5]`. Generally you want as little symmetry as possible so that you can catch errors that just transpose indices.

gridData/tests/test_dx.py
```diff
@@ -90,3 +90,26 @@ def test_delta_precision(tmpdir):
         g.delta, g2.delta,
         decimal=7,
         err_msg="deltas of written grid do not match original")
+
+def test_dx_from_grid():
+    data = np.ones((5, 5, 5), dtype=np.float32)
```
orbeckst: use a non-symmetric grid eg `arange(60).reshape(3,4,5)`

gridData/tests/test_dx.py
```diff
@@ -90,3 +90,26 @@ def test_delta_precision(tmpdir):
...
+    
+    assert isinstance(dx_field, gridData.OpenDX.field)
+    
+    assert any(('test_density' and 'test_user') in str(c) for c in dx_field.comments) 
+    
+    assert_allclose(dx_field.components['data'].array, data)
+    assert_allclose(dx_field.components['positions'].origin, g.origin)
+    
+def test_dx_native():
+    data = np.ones((5, 5, 5))
```
orbeckst: use a non-symmetric grid eg `arange(60).reshape(3,4,5)`

gridData/tests/test_dx.py
```diff
@@ -90,3 +90,26 @@ def test_delta_precision(tmpdir):
...
+    g = Grid(data, origin=[0, 0, 0], delta=[1, 1, 1])
```
orbeckst: vary delta

gridData/tests/test_mrc.py
```diff
@@ -143,167 +179,191 @@ def test_origin(self, grid, ccp4data):
     def test_data(self, grid, ccp4data):
         assert_allclose(grid.grid, ccp4data.array)
 
+    def test_mrc_from_grid(self):
+        data = np.arange(27, dtype=np.float32).reshape((3, 3, 3))
```
orbeckst: even better: use a non-symmetric grid eg `arange(60).reshape(3,4,5)`

gridData/tests/test_mrc.py
```diff
@@ -143,167 +179,191 @@ def test_origin(self, grid, ccp4data):
...
+        g = Grid(data, origin=[1.0, 2.0, 3.0], delta=[0.5, 0.5, 0.5])
```
orbeckst: vary delta, e.g. `[0.5, 1.5, 1.0]`

orbeckst (changes requested):
Good that you got it to work! I would like to try and get MRC to work with a stream and avoid a temp file that is not cleaned up. Please see comments.

Tests need a bit more work and the coverage is also below threshold. Have a look at the coverage report to see which lines are not tested.

Good work!

orbeckst:
Regarding coverage: Just check that the lines that _you wrote_ get covered. You ran black on the whole file so this picks up any of those lines. You are not responsible for writing other tests if this was just a formatting change (this is one of the reasons why we normally do reformatting in separate PRs.)

orbeckst:
Looking at https://app.codecov.io/gh/MDAnalysis/GridDataFormats/pull/164 it seems that the `Grid.convert_to()` method is not tested. This one _is_ important :-) so make sure there are explicit tests for OpenDX and MRC.

gridData/tests/test_mrc.py
spyke7: If I do the native test with `np.ones`  and symmetric matrix, then it will be no problem whether I give data or data.T
In case of (3,4,5) if I don't take the transpose it is throwing an AssertionError
as (x,y,z) -> (z,y,x)

the error - 
```
 AssertionError:
E       Not equal to tolerance rtol=1e-07, atol=0
E
E       (shapes (5, 4, 3), (3, 4, 5) mismatch)
```

spyke7: dx is working fine with just checking with `data`, with non-symmetric and varying delta

spyke7: https://mrcfile.readthedocs.io/en/stable/source/mrcfile.html#mrcfile.new_mmap

Check for shape

gridData/mrc.py
spyke7: Thanks for the suggestions,
What I am doing is that inside of `write` I will check whether a stream is given or not and then inside it only, call the `MrcInterpreter`

spyke7:
So, my plan is this - 
- check whether the passed filename is `io.BytesIO` or not using `isinstance`
- create a header with numpy using the header data type
- that header will be given values  - nx, ny, nz, mode, map, machst and then the inside the filename(i.e the stream) the header and the `data_for_file` will be written
- else all other things same for `write`

[mrcfile utils](https://mrcfile.readthedocs.io/en/stable/source/mrcfile.html#module-mrcfile.utils)

problem is - 
from a `mrcfile.mrcinterpreter.MrcInterpreter` object we need to convert it to `mrcfile.mrcfile.Mrcfile`
for that i guess, we need to tweak values inside `__init__` of `mrcfile.mrcfile.Mrcfile`. That way we can convert it.

There is no other way of writing a stream except the interpreter. It would be helpful if I get some info regarding the Mrcfile codebase @orbeckst @BradyAJohnston pls see once when you have got time
Thanks

orbeckst:
I updated with how I think this could work (`git pull` the changes). The tests passed locally. Please have a quick look and let me know what you think.

gridData/tests/test_mrc.py
orbeckst: I now understand that the data inside the mrc object is rearranged 
```python
        data = np.transpose(self.array,
                            axes=inverse_transpose_order).astype(np.float32)
```
compared to the input data and we have to "un-arrange" it. Given that we know the exact input, just add the appropriate reverse operation::
```python
        # invert the internal data representation
        transpose_order = np.array([[0, 1, 2])
        data = np.transpose(self.array,
                            axes=transpose_order).astype(np.float32)
```

spyke7:
> I updated with how I think this could work (`git pull` the changes). The tests passed locally. Please have a quick look and let me know what you think.

Really thanks for the code of Mrcinterpreter. (I didn't have to do much heavylifting 🙏) 
I wasn't able to spot the `_create_default_attributes` func, that is why I told about importing mode, map, etc from the constant and utils. Now the thing is done. And here directly in write() we are giving the _iostream

I was looking if is there any way to still return a Mrcfile object.
I saw into the mrcfile.py and one thing is that if we use `__new__` we could get the raw Mrcfile object and then update the `_mode, _read` and then call the `_create_default_attributes` then we can return the Mrcfile object. 
I have edited the mrc.py and the above thing works.

This is how the `__init__` of mrcfile.py looks - 
```shell
       super(MrcFile, self).__init__(permissive=permissive, **kwargs)
        
        if mode not in ['r', 'r+', 'w+']:
            raise ValueError("Mode '{0}' not supported".format(mode))

        name = str(name)  # in case name is a pathlib Path
        if ('w' in mode and os.path.exists(name) and not overwrite):
            raise ValueError("File '{0}' already exists; set overwrite=True "
                             "to overwrite it".format(name))
        
        self._mode = mode
        self._read_only = (self._mode == 'r')
        
        self._open_file(name)
        
        try:
            if 'w' in mode:
                self._create_default_attributes()
            else:
                self._read(header_only)
        except Exception:
            self._close_file()
            raise
```

But it will require the io.BytesIO() stream. (The idea we talked about)
So the native looks something like this - 
```shell
mrc = mrcfile.mrcfile.MrcFile.__new__(mrcfile.mrcfile.MrcFile)
mrc._iostream = io.BytesIO()
mrc._mode = 'r+'
mrc._read_only = False
mrc.permissive = True

mrc._create_default_attributes()
```

But when I'm doing the isinstance check then it works for both `mrcfile.mrcfile.MrcFile` and `mrcfile.mrcinterpreter.MrcInterpreter` 😅

spyke7:
Plz give any suggestions or thoughts, like should I do this?

orbeckst:
Your approach to make the native object MrcFile is interesting. I don't think we really need to return `MrcFile`, though. There's not that much to be gained by having an underlying IO buffer connected to it and `MrcInterpreter` works just as we need it to and even though the assigning an open stream to `_iostream` is not very elegant, it's how the mrcfile library documents use of the class so it's good enough for us.

orbeckst (changes requested):
Let's keep MrcInterpreter.

Could you please improve the tests as outlined in the comments? I worry that issues are slipping through because we use symmetrical arrays as test inputs. Once that's done (and no other issues come up), the PR is done!

gridData/tests/test_mrc.py
```diff
@@ -1,128 +1,164 @@
...
-    assert_allclose(ccp4data.origin, [-12.5404758,  -2.1809523,  57.151876 ])
+    assert_allclose(ccp4data.origin, [-12.5404758, -2.1809523, 57.151876])
+
 
 def test_triclinic_ValueError():
-    with pytest.raises(ValueError,
-                       match="Only orthorhombic unitcells are currently "
-                       "supported, not"):
+    with pytest.raises(
+        ValueError, match="Only orthorhombic unitcells are currently " "supported, not"
```
orbeckst: reformat
```suggestion
        ValueError, match="Only orthorhombic unitcells are currently supported, not"
```

gridData/tests/test_mrc.py
```diff
@@ -143,167 +179,199 @@ def test_origin(self, grid, ccp4data):
...
+
+    def test_mrc_native_property(self):
+        data = np.arange(60, dtype=np.float32).reshape((3, 4, 5))
+        g = Grid(data, origin=[0, 0, 0], delta=[1, 3, 2.5])
+
+        mrc_wrapper = mrc.MRC.from_grid(g)
+        native_mrc = mrc_wrapper.native
+
+        assert isinstance(native_mrc, mrcfile.mrcinterpreter.MrcInterpreter)
+        np.testing.assert_allclose(native_mrc.data, data.T)
```
orbeckst, line 204: Add a comment for future devs:
```suggestion
        # The default axis order in MRC is the transpose of the grid data:
        np.testing.assert_allclose(native_mrc.data, data.T)
```

orbeckst (approved):
Ready to be merged! Good work!! Thank you for addressing the test issues. (There's some missing coverage but that's not your code.)

I am adding two comments and then I'll merge.

orbeckst:
@BradyAJohnston do you want to have a final look? Anything that would prevent this PR to be merged?

orbeckst:
@spyke7 please add yourself to AUTHORS for this PR as this will become your first contribution to GridDataFormats.

gridData/tests/test_dx.py
```diff
@@ -80,13 +87,53 @@ def test_write_dx_ValueError(tmpdir, nptype, outfile, counts=100, ndim=3):
...
+    g = Grid(data, origin=[0, 0, 0], delta=[1, 1.5, 2.5])
+
+    g.metadata["name"] = "test_density"
+    g.metadata["author"] = "test_user"
+
+    dx_field = gridData.OpenDX.field.from_grid(g)
+
+    assert isinstance(dx_field, gridData.OpenDX.field)
+
+    assert any(("test_density" and "test_user") in str(c) for c in dx_field.comments)
```
BradyAJohnston: This logical assert is still broken. It's changed from [this comment](https://github.com/MDAnalysis/GridDataFormats/pull/164#discussion_r2961998435) but still doesn't work as you expect.


```bash
In [4]: ("test_density" and "test_user") in ["test_density"]
Out[4]: False
```

Because the initial brackets resolves to this:

```bash
In [1]: ("test_density" and "test_user")
Out[1]: 'test_user'
```

Keeping them combined into a single assert is obviously not great for us to avoid complexity, so just split into separate asserts for each string.

```py
assert any("test_density" in str(c) for c in dx_field.comments)                                                                         
assert any("test_user" in str(c) for c in dx_field.comments)  
```

gridData/core.py
```diff
@@ -212,6 +212,12 @@ class Grid(object):
 
     #: Default format for exporting with :meth:`export`.
     default_format = 'DX'
+    
+    converter = {
+        'MRC': mrc.MRC.from_grid, 
+        # 'VDB': None,
```
BradyAJohnston: I think we should just remove this commented out code from the PR. Cleaner that way.

gridData/mrc.py
```diff
@@ -101,6 +101,163 @@ def __init__(self, filename=None, assume_volumetric=False):
...
+        MRC
+            MRC wrapper object
+
+
+        .. versionadded:: 1.2.0
+        """
+        mrc_obj = cls()
+
+        mrc_obj.array = grid.grid
+        mrc_obj.delta = np.diag(grid.delta)
```
BradyAJohnston, line 130: I might be wrong here @orbeckst but a `Grid` is already storing the `grid.delta` as a 3x3 array, so `np.diag(grid.delta)` would return a 1D array? So we should just omit the `np.diag`?

gridData/OpenDX.py
```diff
@@ -511,24 +545,87 @@ def __init__(self,classid='0',components=None,comments=None):
...
+            data=array(3, grid.grid, type=type, typequote=typequote),
+        )
+        dx_field = cls("density", components=components, comments=comments)
+        return dx_field
+
+    @property
+    def native(self):
+        """Return native object
+        
+        The "native" object is the :class:gridData.OpenDX.field itself.
```
BradyAJohnston:
```suggestion
        The "native" object is the :class:`gridData.OpenDX.field` itself.
```

CHANGELOG
```diff
@@ -13,11 +13,19 @@ The rules for this file:
...
- * 1.1.1
+ * 1.2.0
 
- Fixes
+  Enhancements 
+
+  * new API-level interoperability features (#160)
+    * new `Grid.convert_to()` method to create the native object (#161, PR #164)
+    * `mrc.MRC`, `OpenDX.field`: new `from_grid()` class method to create the
+      correpsonding GridDataFormats wrapper from a `Grid`; the new attribute `native`
```
BradyAJohnston:
```suggestion
      corresponding GridDataFormats wrapper from a `Grid`; the new attribute `native`
```

gridData/OpenDX.py
```diff
@@ -1076,7 +1216,9 @@ def __peek(self):
         except IndexError:
             raise DXParserNoTokens
 
-    def __consume(self,):
+    def __consume(
```
BradyAJohnston, line 1242: Is there a reason you are doing `__consume` and not `_consume`? Single underscore is used for private methods, double is usually for dunder methods `__init__` etc.

gridData/mrc.py
```diff
@@ -156,115 +322,45 @@ def edges(self):
...
+            + self.origin[d]
+            - 0.5 * self.delta[d, d]
+            for d in range(self.rank)
+        ]
 
     def histogramdd(self):
         """Return array data as (edges,grid), i.e. a numpy nD histogram."""
-        return (self.array, self.edges)
-
+        return (self.array, self.edges)
```
BradyAJohnston:
```suggestion
        return (self.array, self.edges)
```

BradyAJohnston (changes requested):
Overall looking good, but some minor things that need addressing / answering (`np.diag`, `__comsume`).

gridData/tests/test_dx.py
spyke7: Yep! 👍

gridData/OpenDX.py
spyke7: Hi thanks for pointing it, but the `__consume` was already part of existing module with double underscore.
as I have run `black`, the indentation and spacing might be changed

BradyAJohnston: @orbeckst is there a reason it was named `__consume`? and not `_consume`?

orbeckst: This is the closest to a private method that Python has (it gets name mangled). When I wrote the code _a long_ time ago I must have been of the opinion that it's important that nobody else gets to use it except the code itself (because otherwise it would mess up the parsing). I would leave it as is — it's not part of @spyke7 's code and just got reformatted — and we can do a clean-up elsewhere if needed.

gridData/mrc.py
orbeckst: No, `Grid.delta` is 1D array with 3 elements. See https://www.mdanalysis.org/GridDataFormats/gridData/core.html#gridData.core.Grid.delta

The code here looks correct because mrc delta is a 3x3 matrix https://www.mdanalysis.org/GridDataFormats/gridData/formats/mrc.html#gridData.mrc.MRC.delta

orbeckst:
~~@spyke7 don't forget AUTHORS.~~

EDIT: oops. overlooked that you'd already done it

gridData/OpenDX.py
BradyAJohnston: If it's internal & existing then best to just leave as-is!

orbeckst:
I don't need to review again as you addressed my concerns, so no need to "request review" from me. I'll just let my approving review stand and wait for @BradyAJohnston to look at how you addressed his requests. Once he's happy, I can squash-merge and we can go back to VDB.

gridData/tests/test_dx.py
orbeckst:
```python
In [1]: ("test_density" and "test_user")
Out[1]: 'test_user'
```
Just weird how Python handles boolean operators... I would have guessed it returns `True`. TIL.

spyke7:
I was actually busy this week due to some assignments in college. Is there any other things to change @BradyAJohnston ?

BradyAJohnston (approved):
LGTM

orbeckst:
@spyke7 congratulations, your first major PR was merged into GDF! 🎉 

I think you should now be able to finish PR #148 (together with #163).

PR #165, opened 2026-04-10
fix inconsistent error handling for Grid.load()

orbeckst:
- gOpenMol parser now captures its TypeError and raises ValueError if the header cannot be read; this makes it consistent with the other parsers
- removed XFAILs from tests that check failure behavior for incorrectly specified file_format for Grid/Grid.load()
- update CHANGELOG
- update docs for Grid.load()

### DIFF
--- a/gridData/gOpenMol.py
+++ b/gridData/gOpenMol.py
@@ -209,12 +209,17 @@ class Plt(object):
         from struct import calcsize, unpack
         if not filename is None:
             self.filename = str(filename)
-        with open(self.filename, 'rb') as plt:
-            h = self.header = self._read_header(plt)
-            nentries = h['nx'] * h['ny'] * h['nz']
-            # quick and dirty... slurp it all in one go
-            datafmt = h['bsaflag']+str(nentries)+self._data_bintype
-            a = numpy.array(unpack(datafmt, plt.read(calcsize(datafmt))))
+        try:
+            with open(self.filename, 'rb') as plt:
+                h = self.header = self._read_header(plt)
+                nentries = h['nx'] * h['ny'] * h['nz']
+                # quick and dirty... slurp it all in one go
+                datafmt = h['bsaflag'] + str(nentries) + self._data_bintype
+                a = numpy.array(unpack(datafmt, plt.read(calcsize(datafmt))))
+        except Exception as err:
+            raise ValueError(f"gOpenMol PLT file {filename} could not be read. "
+                             "The error was\n"
+                             f"   {err.__class__.__name__}: {err}")
         self.header['filename'] = self.filename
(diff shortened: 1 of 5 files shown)


gridData/gOpenMol.py
```diff
@@ -209,12 +209,17 @@ def read(self, filename):
...
-            datafmt = h['bsaflag']+str(nentries)+self._data_bintype
-            a = numpy.array(unpack(datafmt, plt.read(calcsize(datafmt))))
+        try:
+            with open(self.filename, 'rb') as plt:
+                h = self.header = self._read_header(plt)
+                nentries = h['nx'] * h['ny'] * h['nz']
+                # quick and dirty... slurp it all in one go
+                datafmt = h['bsaflag'] + str(nentries) + self._data_bintype
+                a = numpy.array(unpack(datafmt, plt.read(calcsize(datafmt))))
+        except Exception as err:
```
BradyAJohnston: Do we want this to be more specific with the exceptions it catches?

gridData/tests/test_grid.py
```diff
@@ -176,15 +176,12 @@ def test_init_pickle_pathobjects(self, data, tmpdir):
...
         assert h == data['grid']
 
-    # At the moment, reading the file with the wrong parser does not give
-    # good error messages.
-    @pytest.mark.xfail
     @pytest.mark.parametrize("fileformat", ("ccp4", "plt", "dx"))
-    def test_load_wrong_fileformat(self, data, pklfile, fileformat):
-        with pytest.raises('ValueError'):
+    def test_load_wrong_fileformat_raises_ValueError(self, pklfile, fileformat):
+        with pytest.raises(ValueError):
```
BradyAJohnston, line 198: maybe add a `match=` to be more specific?

BradyAJohnston (commented):
Minor comments, LGTM

gridData/gOpenMol.py
orbeckst: Sharpened it, I _think_ it will only ever trigger a TypeError

gridData/tests/test_grid.py
orbeckst: The underlying code is not consistent with the error message and can raise a ValueError from different parts (mainly the MRC code). I didn't want to hide ValueErrors that don't come from wrong input files so I left it as is. 

Detecting what went wrong when a Grid fails to load is still fairly messy but I didn't have time to fundamentally change how it's done at the moment. Hopefully "good enough".

orbeckst:
Thanks for the comments @BradyAJohnston . I addressed them at least partially (see explanation in comments). Btw, don't hesitate to leave a "request changes", please.

BradyAJohnston (approved):
LGTM

Issue #162, opened 2026-02-09
API-interoperability: implement loading of a Grid from a native object


orbeckst:
Implement creating a Grid from a native object as described in the parent issue #160 for

- mrcfile
- openvdb (once we have #141 )

orbeckst:
We can create sub-issues for MRC and VDB.

orbeckst:
In principle, we should treat
- filename (str)
- array (array_like) + edges + origin
- grid (Grid)

also as part of how to treat the input argument `grid`.

To make this approach maintainable and extensible, we should be thinking about breaking the loading process into steps
1. format detection
2. dispatch to a loader that returns either a Grid or a wrapper data structure
3. extract data from loader into the Grid (if necessary)

Ideally, we have the loader functionality bundled with the wrapper classes when possible so that all format-specific code lives in the same place.

orbeckst:
Make sure to document the "reading" capabilities for OpenVDB, namely from PR https://github.com/MDAnalysis/GridDataFormats/pull/170
```python
import openvdb as vdb
grids, _ = vdb.readAll(VDBFILE)
grid_vdb = grids[0]
g = Grid(grid=grid_vdb)
```

PR #170, opened 2026-05-29
Added support for loading native object for openvdb

### LINKED ISSUE #171: create Grid from openvdb grid object
Add capability so that `Grid(grid=openvdb.FloatGrid(...))` creates a functional `Grid`.

Any scalar grid should work and the most appropriate numpy dtype should be chosen (do not loose precision, conserve memory).

See #162 and #160 for context.

spyke7:
Fixes #171

Changes - 
- Made `_extract_from_vdb_grid` that gets the delta, origin, data type and metadata, if a native object is passed - OpenVDB.py
- made the datatypes list global as _DATATYPES and checked for instance if a openvdb.gridbase is passed - OpenVDB.py
- In core.py, checked if not a filename (str) is given, and called the OpenVDBField - core.py
- Updated the tests for the `_extract_from_vdb_grid` - test_vdb.py

@orbeckst please have a look at this when you are free and have time.
Also, made a new section in changelog, with enhancement in it


orbeckst:
Is the objective of the PR to enable #162 for openvdb objects only or implement it for _all_ available formats?

It's totally ok to only do it for OpenVDB, I just want to know when reviewing the PR.

CHANGELOG
```diff
@@ -13,6 +13,17 @@ The rules for this file:
   * accompany each entry with github issue/PR number (Issue #xyz)
 
 -------------------------------------------------------------------------------
+??/??/???? orbeckst, spyke7
```
orbeckst:
```suggestion
??/??/???? spyke7
```
Just you, I haven't done anything yet on the next release ;-)

CHANGELOG
```diff
@@ -13,6 +13,17 @@ The rules for this file:
...
+
+  * 1.2.1
```
orbeckst: Under [semantic versioning](https://semver.org) we have to increase the minor
```suggestion
  * 1.3.0
```

gridData/OpenVDB.py
```diff
@@ -272,6 +294,57 @@ def native(self):
...
+    def _extract_from_vdb_grid(self):
+        """Extract numpy array, origin, delta from stored VDB grid.
+
+        This method converts the sparse VDB grid to a dense numpy array
+        and extracts the transform information.
+        """
+        for key in self.vdb_grid.metadata:
+            try:
+                self.metadata[key] = self.vdb_grid[key]
+            except (TypeError, ValueError):
```
orbeckst: Under which conditions does that fail? I am always suspicious of try/except that passes.

gridData/OpenVDB.py
```diff
@@ -272,6 +294,57 @@ def native(self):
...
+
+        transformation = self.vdb_grid.transform
+
+        v_origin = np.array(transformation.indexToWorld([0, 0, 0]))
+        v_delta = np.array(transformation.indexToWorld([1, 1, 1])) - v_origin
+
+        self.origin = v_origin
+        self.delta = v_delta
+
+        dtype = np.dtype("float32")
```
orbeckst: I think it's cleaner to set the dtype to the default in an `else` block of the `for` loop:
```python
        for numpy_dtype, vdb_names in self._DATATYPES.items():
            name_dtype = vdb_names[0]
            ...
            if vdb_class_name == canonical_name:
                dtype = numpy_dtype
                break
        else:
             # could not find a matching dtype, use default
             dtype = np.float32
```

Or maybe should we use float64? @BradyAJohnston @PardhavMaradani do you have an opinion?

orbeckst: You should also log a warning if you encounter an unknown type and use the default.

gridData/OpenVDB.py
```diff
@@ -272,6 +294,57 @@ def native(self):
...
+            name_dtype = vdb_names[0]
+            canonical_name = (
+                name_dtype.gridType
+                if isinstance(name_dtype, DownCastTo)
+                else name_dtype
+            )
+
+            if vdb_class_name == canonical_name:
+                dtype = numpy_dtype
+                break
```
orbeckst: The logic is flawed. You are picking the first match to a value in the `_DATATYPES` dict and the relationship from vdb to np is many-to-one (eg FloatGrid can map back to int8, unint16, float32, ...) and the dict order is not defined so you might choose the wrong dtype.

Instead you _must_ define an appropriate mapping from VDB to np.

gridData/OpenVDB.py
```diff
@@ -154,6 +155,22 @@ class OpenVDBField(object):
...
+        np.dtype("int16"): ["Int32Grid", "FloatGrid"],
+        np.dtype("uint16"): ["Int32Grid", "FloatGrid"],
+        np.dtype("int32"): ["Int32Grid", DownCastTo("FloatGrid")],
+        np.dtype("uint32"): [DownCastTo("Int32Grid"), DownCastTo("FloatGrid")],
+        np.dtype("int64"): ["Int64Grid", DownCastTo("FloatGrid")],
+        np.dtype("uint64"): ["Int64Grid", DownCastTo("FloatGrid")],
+        np.dtype("float16"): ["HalfGrid", "FloatGrid"],
+        np.dtype("float32"): ["FloatGrid"],
+        np.dtype("float64"): ["DoubleGrid", DownCastTo("FloatGrid")],
+    }
```
orbeckst: I like moving the dict into the class.

However, you _cannot use this dict for the mapping from openvdb to np_, you have to create a separate one. (See comments on the code that currently uses this dict.) It can be as simple as
```python
_DTYPES_VDB2NP = {"BoolGrid": np.dtype("bool"),
              "Int32Grid": np.dtype("int32"),
              ...
              "DoubleGrid": np.dtype("float64")",
             }
```
In this direction there's a clear one-to-one mapping and no downcast needed.

gridData/OpenVDB.py
```diff
@@ -154,6 +155,22 @@ class OpenVDBField(object):
       g.export('output.vdb', format='vdb')
     """
 
+    # dtype maps
+    _DATATYPES = {
```
orbeckst: Perhaps rename, eg `_DTYPES_NP2VDB` ??

gridData/OpenVDB.py
```diff
@@ -272,6 +294,57 @@ def native(self):
...
+        bbox = self.vdb_grid.evalActiveVoxelBoundingBox()
+
+        if bbox is None or bbox[0] == bbox[1]:
+            self.grid = np.zeros((0, 0, 0), dtype=dtype)
+            return
+
+        shape = tuple(np.array(bbox[1]) - np.array(bbox[0]) + 1)
+
+        self.grid = np.zeros(shape, dtype=dtype)
+        print(dtype)
```
orbeckst, line 350: remove

gridData/OpenVDB.py
```diff
@@ -272,6 +294,57 @@ def native(self):
...
+
+            if vdb_class_name == canonical_name:
+                dtype = numpy_dtype
+                break
+
+        bbox = self.vdb_grid.evalActiveVoxelBoundingBox()
+
+        if bbox is None or bbox[0] == bbox[1]:
+            self.grid = np.zeros((0, 0, 0), dtype=dtype)
+            return
```
orbeckst: What happens here?

Possibly log a warning that you encountered an empty box with a single entry. (Is that what happened??)

gridData/core.py
```diff
@@ -249,6 +249,21 @@ def __init__(self, grid=None, edges=None, origin=None, delta=None,
...
+                    self.metadata = vdb_field.metadata
+                    self._load(
+                        grid=vdb_field.grid,
+                        origin=vdb_field.origin,
+                        delta=vdb_field.delta,
+                        metadata=vdb_field.metadata,
+                    )
+                    return
+            except ImportError:
+                pass
```
orbeckst, line 266: This is the right idea. This (together with the filename/array code below) is quite clunky. For each additional format we will need to add another try/except. 

We need to come up with a more general solution, e.g., a dict-based dispatch lookup or  a dispatcher function that calls the correct setup, based on the detected format.

gridData/tests/test_vdb.py
```diff
@@ -392,6 +392,20 @@ def test_from_native_grid_shape_values_and_dimension(self, grid345):
         world = native_grid.transform.indexToWorld((0, 0, 0))
         assert_allclose([world[0], world[1], world[2]], g.origin, rtol=1e-5)
 
+    def test_extract_from_vdb_grid(self, grid345):
```
orbeckst, line 414: parameterize the test to try out all normally supported vdb gridtypes

gridData/tests/test_vdb.py
```diff
@@ -392,6 +392,20 @@ def test_from_native_grid_shape_values_and_dimension(self, grid345):
...
+        native = g.convert_to("vdb")
+        new_vdb_grid = Grid(grid=native)
+
+        assert_allclose(new_vdb_grid.grid, g.grid, rtol=1e-5)
+        assert_allclose(new_vdb_grid.origin, g.origin, rtol=1e-5)
+        assert_allclose(new_vdb_grid.delta, g.delta, rtol=1e-5)
+        assert new_vdb_grid.metadata["name"] == "new_density"
+        assert new_vdb_grid.grid.dtype == np.dtype("float32")
+        assert_allclose(new_vdb_grid.grid, data, rtol=1e-5)
+
```
orbeckst, line 427: Also include a test that shows that file round-tripping works, e.g., 
1. create `Grid`
2. write OpenVDB
3. read OpenVDB
4. compare to `Grid`

orbeckst: Include a test that reads the test vdb file directly with OpenVDB first.

orbeckst (changes requested):
Great that you're starting the work on #162 . The type-selection mechanism needs to be reworked. 

We also need to have a more general architecture for `Grid` for (1) detecting the format, (2) dispatching to the correct loader. We can decide to do this in a separate PR if for right now we're only focusing on VDB.

orbeckst:
For right now I linked this PR to the OpenVDB-only issue #171

gridData/OpenVDB.py
spyke7: yep, I will try to redo that

gridData/OpenVDB.py
spyke7: basically for VDB grid that has no active voxels
evalActiveVoxelBoundingBox() returns none

orbeckst: Thanks for the explanation. Can you add a comment to the code that explains what happens?

And issue a warning.

gridData/OpenVDB.py
spyke7: vec3f can be one of the example. But I can create a path for a warning to be raised for this

gridData/OpenVDB.py
spyke7: Please once relook at the logic

gridData/core.py
spyke7: jus like `converter`, can map different loader for native in a dict
But before that need to check the type of grid passed, then Ig it will look clean.

gridData/OpenVDB.py
```diff
@@ -272,6 +304,58 @@ def native(self):
...
+        self.origin = v_origin
+        self.delta = v_delta
+
+        dtype = self._DTYPES_VDB2NP.get(type(self.vdb_grid).__name__)
+        if dtype is None:
+            warnings.warn(
+                f"Unknown VDB grid type '{type(self.vdb_grid).__name__}', defaulting to float32.",
+                RuntimeWarning,
+            )
+            dtype = np.dtype("float32")
```
orbeckst: This looks correct but I'd probably write it as
```python
try:
   dtype = self._DTYPES_VDB2NP[type(self.vdb_grid).__name__]
except KeyError:
   warning(...)
   dtype = np.dtype("float32")
```

gridData/OpenVDB.py
orbeckst: still a print there!

gridData/OpenVDB.py
orbeckst: Actually, fail here and raise a TypeError exception.

Better to fail than to try a weird conversion. For instance, we have no idea what to do with a VecGrid (or whatever it's called).

spyke7:
Hi Prof @orbeckst 
I was busy with my sem exams and going on, that's why I wasn't active regarding this. I hope, I can apply more improvements and changes. 
Thanks

orbeckst (approved):
Thanks for making the last changes. This looks ready, nice work!

Apologies for the delayed review.

spyke7:
> Thanks for making the last changes. This looks ready, nice work!
> 
> Apologies for the delayed review.

Thanks professor @orbeckst . Learned a lot about openvdb and the other APIs during this!! 😊

PR #179, opened 2026-07-08
fixed initialization of DXClasses

orbeckst:
- updated init signature of OpenDX.DXClass to include name and components
- consistently use super() in DXClass-child classes
- use zero-arg super() calls
- see https://github.com/MDAnalysis/GridDataFormats/security/quality/rules/py%2Fmissing-call-to-init


orbeckst (commented):
everything passes, lgtm

